PR #3, opened 2015-10-13
Pull functionality out into more specific modules.

DaveShuckerow:
Detailed changes:
- create database module.  This will allow us to change the database with sqlalchemy without having to spin up the server.
- create util module.  This contains utility functions and entities -- bcrypt, login management, and functions for serving static content, templated html and the like can live here.
- fix the dirty static html hack from stack overflow to use a builtin flask function.
- move login function to the user models.py module.  We had to import that module in the middle of the **init** code to get it working there, and since its purpose is to log a user in from the database, this seems like a more reasonable home for it.

### DIFF
--- a/auacm/app/database.py
+++ b/auacm/app/database.py
@@ -1,0 +1,11 @@
+'''Database handlers'''
+from sqlalchemy import create_engine
+from sqlalchemy.orm import Session
+from sqlalchemy.ext.automap import automap_base
+
+# database setup
+Base = automap_base()
+engine = create_engine('mysql://acm@localhost/acm')
+connection = engine.connect()
+Base.prepare(engine, reflect=True)
+session = Session(engine)
--- a/auacm/app/database.py
+++ b/auacm/app/database.py
@@ -1,4 +1,4 @@
-'''Database handlers'''
+'''Database handlers.'''
 from sqlalchemy import create_engine
 from sqlalchemy.orm import Session
 from sqlalchemy.ext.automap import automap_base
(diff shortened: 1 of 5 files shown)


auacm/app/util.py
```diff
@@ -0,0 +1,29 @@
...
+# bcrypt setup
+bcrypt = Bcrypt(app)
+
+
+# login session setup
+login_manager = LoginManager()
+login_manager.init_app(app)
+
+
+def serve_html(filename):
```
DaveShuckerow: Here's the cleaned-up way of serving static html I mentioned.

auacm/app/util.py
```diff
@@ -0,0 +1,29 @@
...
+    '''Serve static HTML pages.'''
+    return send_from_directory(app.static_folder+"/html/", filename)
+
```
DaveShuckerow, line 20: Here's the cleaned-up way of serving static html that I mentioned.

WilliamHester, line 20: I like it. Much better than the SO implementation I had before.

auacm/app/database.py
```diff
@@ -0,0 +1,11 @@
+'''Database handlers.'''
+from sqlalchemy import create_engine
+from sqlalchemy.orm import Session
+from sqlalchemy.ext.automap import automap_base
+
+# database setup
+Base = automap_base()
+engine = create_engine('mysql://acm@localhost/acm')
```
WilliamHester, line 8: We might want to abstract this out into Config.py in the event that someone's mysql server doesn't have the account "acm" set up with the ACM database.

DaveShuckerow, line 8: I think pulling away the account dependency is a good idea, but I'm not entirely sure that this is the right way to go about it yet.  Because config.py isn't checked into the repo, we need another way of providing default options.  I think the ideal approach would be to add this to config.py, but to also add this to setup scripts that can configure the default option.

PR #5, opened 2015-10-23
Blog

WilliamHester:
It's blog time.


DaveShuckerow:
Would it be better to just have the wordpress server underneath the urls for the blog?

WilliamHester:
This is approximately 1000x more lightweight than wordpress and allows us to be in full control. Wordpress also has some security problems that I don't feel like dealing with. If there's a real reason to move to Wordpress in the future, we can definitely convert to it, but I'd rather not for now.

DaveShuckerow:
OK.

PR #6, opened 2015-10-23
create-user

BrandonLMorris:
Add functionality for admins to create new users


auacm/app/static/js/controllers/SettingsController.js
```diff
@@ -33,4 +34,44 @@ app.controller('SettingsController', ['$scope', '$http', function($scope, $http)
...
+            transformRequest: function(obj) {
+                var str = [];
+                for (var p in obj)
+                    str.push(encodeURIComponent(p) + "=" + encodeURIComponent(obj[p]));
+                console.log("Request: " + str.join("&"));
+                return str.join("&");
+            },
+            data: request
+        }).then(function(response) {
+            if (response.status == 200) {
```
WilliamHester: This is not necessary. The first function is called if the status code is from [200 - 299]. [See this.](https://docs.angularjs.org/api/ng/service/$http)

WilliamHester: Also, after a success, let's clear the fields so that the user doesn't have to manually clear them when creating multiple new users at a time.

auacm/app/static/js/controllers/SettingsController.js
```diff
@@ -1,5 +1,6 @@
 app.controller('SettingsController', ['$scope', '$http', function($scope, $http) {
     $scope.changePassword = function() {
+        console.log($scope.newPassword);
```
WilliamHester: Looks like you forgot to take this out.

auacm/app/static/js/controllers/MainController.js
```diff
@@ -1,18 +1,35 @@
...
         },
         function(error) {
-            
+
         });
     $scope.signOut = function() {
         // Here, we have the username and password, accessible by
         //     $scope.username and $scope.password. We need to call
         //     the backend to log in.
+        // ???
```
WilliamHester: I remove this accidental comment in the blog branch.

auacm/app/static/js/controllers/MainController.js
```diff
@@ -1,18 +1,35 @@
 app.controller('MainController', ['$scope', '$http', '$route', '$window', 
         function($scope, $http, $route, $window) {
-    $scope.username = 'placeholder'
-    $scope.$route = $route
+    // Intialize user fields
+    $scope.username     = 'placeholder';
+    $scope.displayName  = 'placeholder';
+    $scope.isAdmin      = false;
+
+    $scope.$route       = $route;
```
WilliamHester, line 8: Weird spacing here, but it's okay I guess.

auacm/app/static/js/controllers/SettingsController.js
BrandonLMorris: Ok, I was following your example from lines 25-35. Should I change those too? Is there every an instance when the status code would be 201-299 and we would need to report something wrong?

auacm/app/static/js/controllers/MainController.js
BrandonLMorris, line 8: It lines up the ='s. It looks nice, but it is inconsistent.

PR #8, opened 2015-10-25
Judge Tests

DaveShuckerow:
This is a big one.  
- I refactored the judge code out of the flask code into a separate file to make it more testable.
- I wrote unit tests for getting different execution results on a Java submission

Still TODO(for another pull request):
- support for both python2 and python3 submissions (based on first-line comment)
- test the messages the judge submissions emit
- test to make sure that new languages have tests verifying that they work _before_ we start using them.


auacm/app/modules/submission_manager/judge.py
```diff
@@ -0,0 +1,225 @@
...
+    "go": "go build -o {0} {0}.go"
+}
+RUN_COMMAND = {
+    "java": "java -cp {0}/ {1}",
+    "py": "python {0}/{1}.py",
+    "c": "{0}/{1}",
+    "cpp": "{0}/{1}",
+    "go": "{0}/{1}"
+}
+TIMEOUT_MULTIPLIER = {
```
BrandonLMorris, line 27: This works for now, but we've talked about doing empirical tests to determine multipliers

auacm/app/modules/submission_manager/judge.py
```diff
@@ -0,0 +1,225 @@
...
+                 open(os.path.join(result_path, out_file)) as submission_result:
+                golden_lines = golden_result.readlines()
+                submission_lines = submission_result.readlines()
+                if len(submission_lines) != len(golden_lines):
+                    submission.update_status("wrong")
+                    submission.emit_status("incorrect", test_number)
+                    return WRONG_ANSWER
+                # Use itertools.izip instead of zip to save memory.
+                for gl, sl in itertools.izip(golden_lines, submission_lines):
+                    if gl.rstrip() != sl.rstrip():
```
BrandonLMorris: Could this strip trailing whitespace in a submission? ICPC considers trailing whitespace to be incorrect, even if the the stripped versions are equal.

WilliamHester: Yeah, we previously had rstrip('\r\n') which removed only trailing newline characters (correcting for unix/windows incompatibility). I believe we should keep it that way.

auacm/app/modules/submission_manager/judge.py
```diff
@@ -0,0 +1,225 @@
...
+                submission.update_status("timeout")
+                submission.emit_status("timeout", test_number)
+                return TIMELIMIT_EXCEEDED
+            elif execution.process.poll() != 0:
+                submission.update_status("runtime")
+                submission.emit_status("runtime", test_number)
+                return RUNTIME_ERROR
+            result_path = os.path.join(submission_directory, "out")
+            # The execution is completed.  Check its correctness.
+            with open(os.path.join(output_path, out_file)) as golden_result, \
```
WilliamHester: golden_result isn't really descriptive. Let's use something more descriptive like correct_result or judge_result.

auacm/app/modules/submission_manager/judge.py
```diff
@@ -0,0 +1,225 @@
...
+    )
+    if result == 0:
+        return COMPILATION_SUCCESS
+    else:
+        submission.update_status('compile')
+        submission.emit_status("compile", -1)
+        return COMPILATION_ERROR
+
+
+def execute_submission(submission, uploaded_file):
```
WilliamHester, line 92: Vertical whitespace is nice. Since this method is kind of a monster, it would be nice if we could get some vertical whitespace separating the logical breaks.

auacm/app/modules/submission_manager/judge.py
```diff
@@ -0,0 +1,225 @@
...
+    "go": "{0}/{1}"
+}
+TIMEOUT_MULTIPLIER = {
+    "java": 1.5,
+    "py": 2,
+    "c": 1,
+    "cpp": 1,
+    "go": 1
+}
+PENDING = 0
```
WilliamHester: Not a big deal, but PENDING is never used.

auacm/app/modules/submission_manager/views.py
```diff
@@ -48,191 +28,23 @@ def submit():
...
-                         result='start')
-    session.add(attempt)
-    session.flush()
-    session.commit()
-    session.refresh(attempt)
-    directory = directory_for_submission(attempt)
-    os.mkdir(directory)
-    uploaded_file.save(join(directory, uploaded_file.filename))
-    thread = Thread(target=start_execution, args=(attempt, uploaded_file))
+    attempt = models.Submission(
```
WilliamHester, line 36: It's cool if you want to refactor the code, (and I like this style better than the one that I initially used) but in the future, let's avoid changing style during commits. It's not truly productive and can end being a battle.

auacm/app/modules/submission_manager/views.py
```diff
@@ -48,191 +28,23 @@ def submit():
...
+        file_type=uploaded_file.filename.split('.')[-1].lower(),
+        result='start')
+    attempt.commit_to_session()
+    thread = Thread(
+        target=judge.evaluate, args=(attempt, uploaded_file))
+    thread.daemon = False
     thread.start()
     return serve_response({
-        'submissionId' : attempt.job
+        'submissionId': attempt.job
```
WilliamHester, line 49: See above.

auacm/app/modules/submission_manager/judge.py
WilliamHester, line 27: +1. These should be generated in the future by the host machine. Definitely not a concern of this pull request though.

WilliamHester:
Cool, so I have a couple overall comments on the commit. I know it hasn't been explicitly stated anywhere, but since this pull is fairly large, we need to start enforcing some rules. I want to follow [Google's Style Guide for Python](https://google-styleguide.googlecode.com/svn/trunk/pyguide.html) and [Google's Style Guide for JavaScript](https://google.github.io/styleguide/javascriptguide.xml). (We're going to be working on the wiki here with a page describing the project's style and how to contribute, etc.)

Because of this, there are a few changes that I would like you to make. 
1. Doc strings should use double quotes instead of single quotes.
2. The string literals in the code seemingly randomly alternate between single quotes and double quotes. Personally, I prefer single quotes, as they are kind of a middle ground between JavaScript and Python's style guides, as single quotes are preferred in JavaScript.

Also, I don't know what happened with your commits (some of them seem to be duplicated?), but you might consider squashing these commits into a few smaller ones that would make more sense when we look back at the commit history. If you can't that's fine, but in the future, making such small commits is really unnecessary.

auacm/app/modules/submission_manager/views.py
DaveShuckerow, line 36: Didn't mean to step on any toes; I touched this as part of the refactoring and the linter objected to the line lengths.

auacm/app/modules/submission_manager/judge.py
DaveShuckerow: No, that is a big deal -- there's no point in bloating a codebase with objects that it will never use.  This is the sort of thing that makes code review a great tool.

auacm/app/modules/submission_manager/judge.py
DaveShuckerow, line 92: Done.

auacm/app/modules/submission_manager/judge.py
DaveShuckerow: Done.

Also, for future reference, the inspiration is https://en.wikipedia.org/wiki/Gold_standard_(disambiguation)
Golden tests in computing are tests where the results are graded against some ground-truth collection of data, like we have here.  Goldens can be useful for integration tests or "polish" tests -- you can test a frontend's API calls against the backend code with mock data and golden results, or (especially in the world of apps -- it's harder to do this on the web) you can test the appearance of every page to be sure that you don't accidentally break your UI in a section of the app that you never visit.

auacm/app/modules/submission_manager/judge.py
DaveShuckerow: Done.

DaveShuckerow:
Strings are changed.  The differences in my files was places in the refactoring where I copied lines out from other files and was too ~~tired and lazy~~ busy to go back and notice while I was getting this cleaned up, and the linter on my personal computer didn't call me out on it this weekend.

On the subject of style guides, an excellent tool is the pep8 linter for Python.  You can install it with pip.  It's not as strict as Google's style guide (however, it doesn't enforce 2-space tabs or apparently file-level uniformity in quotes)

I'm not entirely sure how some of the commits got duplicated either (it happened while migrating from my work computer to my personal computer and changing the email address on my personal computer's git globals).  I didn't know git supported squashing commits together; I'll look into that.

auacm/app/modules/submission_manager/views.py
WilliamHester, line 36: All good. Just wanted to make sure that we don't go around refactoring people's code just for style.

PR #9, opened 2015-10-26
Improve dependencies

BrandonLMorris:
Fixed a typo from the "No localhost" merge.

Modified the dependencies obtained in index.html to be entirely static (for offline development) with commented out versions of CDN requests below (to be used in production)


BrandonLMorris:
Currently, switching from development to production dependencies consists of adding/deleting a space before the comment blocks. In the future, can use Flask environment variable and let template do the logic.

PR #11, opened 2015-10-30
Fix problems with scoreboard.

WilliamHester:
The scoreboard now updates as it should. It also disconnects from the socket at the end of the competition. The remaining time is also hidden if the competition is over. The problem where the scoreboard would update a problem in a competition that had ended is also gone.

### DIFF
--- a/auacm/app/modules/submission_manager/views.py
+++ b/auacm/app/modules/submission_manager/views.py
@@ -35,7 +35,7 @@ def submit():
 
     attempt = models.Submission(
         username=current_user.username,
-        pid=request.form['pid'].lower(),
+        pid=request.form['pid'],
         submit_time=int(time.time()),
         auto_id=0,
         file_type=uploaded_file.filename.split('.')[-1].lower(),
--- a/auacm/app/static/js/controllers/ScoreboardController.js
+++ b/auacm/app/static/js/controllers/ScoreboardController.js
@@ -14,6 +14,7 @@ app.controller('ScoreboardController', ['$scope', '$http', '$routeParams', '$win
             $scope.compProblems = response.data.data.compProblems;
             $scope.teams = response.data.data.teams;
             genScoreboard($scope);
+            connectToSocket($window, $scope);
 
             // TODO(brandonlmorris) Add offset to server/client times
             var thisTime = Math.floor(new Date().getTime() / 1000);
@@ -22,7 +23,6 @@ app.controller('ScoreboardController', ['$scope', '$http', '$routeParams', '$win
                 $scope.active = true;
                 $scope.timeLeft = $scope.competition.startTime + $scope.competition.length
                         - thisTime;
-                connectToSocket($window, $scope);
                 var timer = $interval(function() {
                     $scope.timeLeft = $scope.competition.startTime + $scope.competition.length 
                             - thisTime;
@@ -31,7 +31,6 @@ app.controller('ScoreboardController', ['$scope', '$http', '$routeParams', '$win
                         $scope.active = false;
                         $scope.timeLeft = 0;
                         $interval.cancel(timer);
-                        $scope.socket.removeAllListeners();
                     }
                 }, 1000);
             }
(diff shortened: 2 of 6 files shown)


DaveShuckerow:
How does this handle solutions that are submitted before the end of the competition but finish being judged after the competition ends?

WilliamHester:
That's a good point. I guess we should always leave the websocket connected, even if the competition is over. It really can't hurt. That means I need to modify the logic to reject submissions that were after the end time of the competition.

DaveShuckerow:
If the submission is associated with the competition, you could ignore any submissions that come in after the end.

Also, a thought: If we keep a queue of submissions that are waiting to be judged, we can send a message, when the queue is empty and the competition over, to the client that indicates the end of the competition.  Any client that connects after that would just get the EOC message and disconnect (or maybe not even attempt to open the websocket)

WilliamHester:
I like this suggestion. Going to do it right now.

Edit: I take that back. Never mind. Too much work for now.

auacm/app/static/js/controllers/ScoreboardController.js
```diff
@@ -61,38 +13,95 @@ app.controller('ScoreboardController', ['$scope', '$http', '$routeParams', '$win
...
             
         });
+}]);
 
-
-    // Open the socket connection
-    var socket = io.connect('http://' + $window.location.host + '/judge')
-    socket.on('connect', function() {
-        console.log('connected');
+var genScoreboard = function($scope) {
```
BrandonLMorris: Is there a particular reason this got moved outside the controller? I only ask to know if this is a pattern we should use in the future.

auacm/app/modules/submission_manager/views.py
```diff
@@ -35,7 +35,7 @@ def submit():
 
     attempt = models.Submission(
         username=current_user.username,
-        pid=request.form['pid'].lower(),
+        pid=request.form['pid'],
```
BrandonLMorris, line 38: So the database uses camel case for the pid's, but our data folders that hold judge input/output are all lowercase, which is probably the source of this bug. When we start using the website to upload new problems (what I'm working on right now), we should probably look to staying consistent by switching the directory names to mixed case.

auacm/app/static/js/controllers/ScoreboardController.js
```diff
@@ -61,38 +13,95 @@ app.controller('ScoreboardController', ['$scope', '$http', '$routeParams', '$win
...
     var viewed = [];
     socket.on('status', function(event) {
-        if (viewed.indexOf(event.submissionId) > -1 || event.status == 'running') {
-            // The scoreboard doesn't care about the problem if it's not done or
-            // if we've already seen it.
+        if (viewed.indexOf(event.submissionId) > -1 || event.status == 'running'
+                || $scope.compProblems.indexOf(event.problemId) < 0
+                || event.submitTime > $scope.competition.startTime + $scope.competition.length) {
+            // The scoreboard ignores the problem for any of the following reasons:
+            // The submission has already been handled, the event's status is running,
```
BrandonLMorris: I know Kattis will update a scoreboard to show a problem is in the process of being judged. Do we want to consider this in future update? I don't know if ICPC does this.

auacm/app/static/js/controllers/ScoreboardController.js
```diff
@@ -61,38 +13,95 @@ app.controller('ScoreboardController', ['$scope', '$http', '$routeParams', '$win
...
+        prevTime = team.time;
+    }
+
+    // Can break if an $apply is already in progress, need to check first (???)
+    if (!$scope.$$phase) {
+        $scope.$digest();
+    }
+}
 
+var connectToSocket = function($window, $scope) {
```
BrandonLMorris: Since it's outside the controller, doesn't it have global scope? That might cause a problem if we want to create a `connectToSocket()` function elsewhere.

BrandonLMorris:
Looks good. I like that we're adding consistency by making all the API functions return time in seconds. Not sure about moving functions outside the controller; my concern is that they have global scope and could interfere with the app in other places. 

At some point, we need to get serious about style: we've been kinda lax about line lengths (Google style implicitly requires 80 characters, from what I understand). And a line comment over a block of code here and there wouldn't be unwelcome...

auacm/app/static/js/controllers/ScoreboardController.js
WilliamHester: I moved it outside of the controller, because I thought it kind of made it look cleaner, but I'm not sure if it's the best route to go. I forgot about scoping, so we don't necessarily want to make these functions available. I'll move them back.

auacm/app/modules/submission_manager/views.py
WilliamHester, line 38: Agreed. We need to make a few modifications to the database and make some rules for how columns need to be laid out.

auacm/app/static/js/controllers/ScoreboardController.js
WilliamHester: Yes. I don't think it'd be too hard, really. Actually, it wouldn't be at all; someone just needs to take the time to do it. :)

auacm/app/static/js/controllers/ScoreboardController.js
WilliamHester: Yeah, good point. I'm going to move these back in because of that.

PR #14, opened 2015-11-21
Problem submit

BrandonLMorris:
Problems are now inCRUDible! The api and web interface now support creating, updating, deleting, and viewing (in HTML) problems. It should be noted that deleting problems has no web interface, and can only be achieved through the api. 

The database was refactored to use integer pid's, converting the old pid's to "shortname". Creating, viewing, and updating problems now have their own pages and Angular controllers.


auacm/app/modules/problem_manager/views.py
```diff
@@ -1,89 +1,341 @@
-from flask import request
-from flask_restful import Resource
-import models
+"""Manages problems within the app, including their creation, deletion,
+updating, and retreival.
+
+Functions:
+url_for_problem -- returns the route for the pdf description of a problem
+                   given an arguemtn that has a 'pid' field.
```
WilliamHester: "arguemtn"

auacm/app/modules/problem_manager/views.py
```diff
@@ -1,89 +1,341 @@
...
+            case_num += 1
+            sample_cases.append(sample)
+
+        in_file = zipfile.ZipFile(request.files['in_file'])
+        out_file = zipfile.ZipFile(request.files['out_file'])
+        sol_file = request.files['sol_file']
+
+    # If any required values were missing, serve an error
+    except KeyError as err:
+        return serve_error('Request header not found: ' + err[0],
```
WilliamHester: Technically, it's not a request header (that would imply that it's an HTTP header). I'd just say something like "'Error, form incomplete; missing ' + err[0]." Something along those lines.

auacm/app/modules/problem_manager/views.py
```diff
@@ -1,89 +1,341 @@
...
+    # Store the judge data
+    directory = os.path.join(app.config['DATA_FOLDER'],
+                             'problems', str(problem.pid))
+    in_file.extractall(directory)
+    out_file.extractall(directory)
+    os.mkdir(os.path.join(directory, 'test'))
+    sol_file.save(os.path.join(directory, 'test', sol_file.filename))
+
+    return serve_response({
+        'success': True,
```
WilliamHester: Why? In the serve_response method, we actually return the response code (200) to verify that the response was good.

auacm/app/modules/problem_manager/views.py
```diff
@@ -1,89 +1,341 @@
...
+        'description': problem_data.description,
+        'input_desc': problem_data.input_desc,
+        'output_desc': problem_data.output_desc,
+        'sample_cases': cases,
+        'pid': problem.pid,
+        'difficulty': problem.difficulty
+    })
+
+# Delete a problem from the database
+@app.route('/api/problems/delete', methods=['POST'])
```
WilliamHester: Ideally, we'd be doing something like

```
@app.route('/api/problems/<pid>', methods=['DELETE'])
```

to actually be a RESTful API.

auacm/app/modules/problem_manager/views.py
```diff
@@ -1,89 +1,341 @@
...
+@login_required
+def delete_problem():
+    """Delete a specified problem in the database and data folder"""
+    # Admin check
+    if not current_user.admin == 1:
+        return serve_error('You must be an admin to delete a problem',
+                           response_code=401)
+
+    if not request.form['pid']:
```
WilliamHester: Let's get this from the route.

auacm/app/modules/problem_manager/views.py
```diff
@@ -1,89 +1,341 @@
...
+    session.flush()
+    session.commit()
+
+    # Delete judge data
+    directory = os.path.join(app.config['DATA_FOLDER'],
+                             'problems', str(request.form['pid']))
+    rmtree(directory)
+
+    return serve_response({
+        'success': True,
```
WilliamHester: Why?

auacm/app/modules/problem_manager/views.py
```diff
@@ -1,89 +1,341 @@
...
+        'deleted_pid': request.form['pid']
+    })
+
+# Update a problem in the database
+@app.route('/api/problems/edit', methods=['POST'])
```
WilliamHester: This should also be using a different HTTP method. Do 

```
@app.route('/api/problems/<pid>', methods=['PUT'])
```

auacm/app/modules/problem_manager/views.py
```diff
@@ -1,89 +1,341 @@
...
-            'name': 'New Problem',
-            'contest': 'Contest name',
-            'difficulty': 'medium'
-        }
-        </code>
-        
-        It returns a <code>Problem</code>
-        """
+    return serve_response({
+        'success': True,
```
WilliamHester: See above.

auacm/app/static/js/modules/mainModule.js
```diff
@@ -54,6 +54,21 @@ app.config(function($routeProvider) {
...
+            templateUrl : 'static/html/createProblem.html',
+            controller : 'CreateProblemController',
+            activetab  : 'settings'
+        })
+        .when('/problems/:pid', { // route for the problem view page
+            templateUrl : 'static/html/viewProblem.html',
+            controller : 'ViewProblemController',
+            activetab : 'problems'
+        })
+        .when('/problems/edit/:pid', { // route for problem edit page
```
WilliamHester: I know this is a little pedantic, but let's do 

```
.when('/problems/:pid/edit', ...
```

to mimic more of a RESTful interface.

WilliamHester:
Slightly more major problem: Bootstrap and Socket.io are not included in the bower install file, so it's partially broken.

WilliamHester:
Looks like sample test cases are dropped when editing a problem. No matter what I do, they seem to disappear.

auacm/app/static/html/createProblem.html
```diff
@@ -0,0 +1,100 @@
...
+          <label for="problemDescription">Problem Description</label>
+          <textarea class="form-control" id="problemDescription" rows="6"
+            ng-model="description"
+            placeholder="Some ridiculous plot..."></textarea>
+        </div>
+
+        <!-- Input Description -->
+        <div class="form-group">
+          <label for="inputDescription">Input Description</label>
+          <textarea class="form-control"
```
WilliamHester: We should bump both of these to at least 3, maybe 4.

WilliamHester:
Looks like competitions are actually broken as well. I'm getting some sort of foreign key problem.

auacm/app/modules/problem_manager/views.py
BrandonLMorris: Why the `'success': True`? I think that was leftover from the original code. I'll gut it.

WilliamHester: You're right. I forgot about the code that I wrote on day 1 with Will that we literally never used.

auacm/app/modules/problem_manager/views.py
BrandonLMorris: Hm, you're right. Should I change (either in this PR or a near future one) the edit action to be like this with a PUT method? I'm not super familiar with REST.

auacm/app/static/html/createProblem.html
BrandonLMorris: Are you talking about the input/output description textareas?

BrandonLMorris:
**On Bootstrap/Socket.io:** I switched to local copies and forgot to tie up the loose ends. I'll add Bootstrap to the init file. However, the (older) version of socket.io we're using I couldn't get from bower. I had to curl it manually. If you know how to do it, I can add it to the init file, otherwise I'll uncomment the CDN request in index.html.

**On sample cases dropping:** Are you supplying sample cases when you edit a problem? Are they just not showing up at all when the edit page loads (they should pre-populate)? If any sample cases are supplied, the old ones are ditched for the new ones; it's all or nothing. It may be mistaking an empty set as a case. I'll look into it.

**On competitions being broken:** How so? I had to add a foreign key field so SQLAlchemy could connect the tables to get the shortnames. I may have forgotten to do another db dump afterword. I'll look into it.

WilliamHester:
**For Bootstrap/Socket.io:** Yeah, if you could add the dependency to the init file, that'd be awesome so we don't have to do that manually. As for socket.io, I say we just leave the CDN in; for offline local development, you can use the curl way.

**For sample cases dropping:** It looks to me like they're just dropping no matter what I do. They're pre-populating just fine, but they're just going away after I save no matter what I do. I'll look more into it to see if it's possible to change/add them at all.

**For competitions:** I'm getting the following error:

```
InvalidRequestError: Could not find a FROM clause to join from.  Tried joining to <class 'app.modules.problem_manager.models.Problem'>, but got: Can't find any foreign key relationships between 'comp_problems' and 'problems'.
```

which traces back to the `.join(Problem)` call on line 51 of views.py in scoreboard_manager. It's probably just the dump or something.

BrandonLMorris:
All corrections should be made now:
- Typo in docstring
- API routes are now more RESTful: editing uses PUT, deletion uses DELETE, and problems are always identified by the route now
- Bootstrap and Socket.io have been added to bower.json
- Database dump (acm.sql) now has the foreign key in comp_problems table so compeition scoreboards will render correctly
- `'success': True` instances have been removed

That should be everything. Web app changes were also made to reflect the api changes, but its (very) possible I missed something.

auacm/app/static/html/createProblem.html
```diff
@@ -0,0 +1,100 @@
...
+            <label for="outputUpload">Judge Output</label>
+            <input type="file" file-model="outFile" accept=".zip">
+            <p class="help-block">All files should be in an 'out' directory and zipped</p>
+          </div>
+        </div>
+
+        <!-- Judge Solution -->
+        <div class="form-group">
+          <label for="judgeSolution">Judge Solution</label>
+          <input type="file" file-model="solFile" accept=".c,.cpp,.java.py">
```
WilliamHester: You accidentally forgot a comma here.

auacm/app/static/html/editProblem.html
```diff
@@ -0,0 +1,100 @@
...
+            <label for="outputUpload">Judge Output</label>
+            <input type="file" file-model="outFile" accept=".zip">
+            <p class="help-block">All files should be in an 'out' directory and zipped</p>
+          </div>
+        </div>
+
+        <!-- Judge Solution -->
+        <div class="form-group">
+          <label for="judgeSolution">Judge Solution</label>
+          <input type="file" file-model="solFile" accept=".c,.cpp,.java.py">
```
WilliamHester: Same story

auacm/app/static/html/createProblem.html
WilliamHester: Yes

WilliamHester:
The new changes are great! It's almost ready to merge, but in trying it out, I found a couple bugs and a couple things that I think should be done a little differently.

**Bugs**
- [x] In both the Edit Problem flow and the Create Problem flow, you can't upload a Java or Python solution. You forgot a comma; no big deal.
- [ ] In the competitions, the problem titles don't line up with the correct submissions*

**General complaints**
- [ ] Edit problems seems essentially to be a clone of create problems, but with different fields. I think we should just merge these two; just figure out what you're doing based off the route or something and load a problem/make a fake one or something.
- [x] After editing/creating a problem, the user should have some sort of confirmation that the problem was successfully created. I'd just do this by changing them to the newly created problem's problem page to show them what they just created.
- [x] I can add empty sample input/output fields just by repeatedly clicking the button.
- Right now there's no form validation, but I actually don't think that needs to be addressed in this pull request.

*on a similar note to form validation, I'll handle the scoreboard stuff since it's kind of a mess, I just wanted to note that it's a bug.

BrandonLMorris:
I concur on merging problem create/edit, however I have two reservations. Combining the front end parts (Angular and HTML) is fairly straightforward; I actually thought about it at the time but I wasn't sure how to go about it, but now I can see how it would work. However, the back end isn't so elegant. I can do it, but it'll basically be putting two method bodies under one signature, not particularly helpful. I'll think on it some more. My other concern is trying to do that in this PR. I'm happy to if its worthwhile, otherwise it'll go to the top of my list (alongside validation).

Just for the record: Concerning the scoreboard display bug, I appreciate you taking care of that. I understand PR's shouldn't introduce bugs, but the case could be made that it was already a bug that just wasn't noticeable. At any rate, thanks for noting it.

Final note: Slap on my wrist for no form validation, I just wanted to get this out of the door. There's also some holes in the API, particularly if a problem is queried that doesn't exist (almost no checks for that). Titles get truncated if they're too long, but that's about it. Whoever takes on validation will need to work on the back end as well.

WilliamHester:
I should have specified. I don't care about the backend as much as the duplicate code on the frontend. I think the backend code actually should be unique (updating vs creating a new thing is inherently different). I'm actually leaning towards merging it and just adding that as an issue that you can work on (since it works, that's just style). It's up to you.

No problem on the scoreboard; I know that should be cleaner.

Understandable; as this is mostly going to be admin-only anyway, I think it's alright for now.

BrandonLMorris:
**Fixed**
- Sample cases can't be added unless the last sample case isn't empty (note that this is just a client-side check)
- Edit/Create now accept Python and Java uploads
- Edit/Create redirect to the problem page upon success

**Still Broken**
- Scoreboard names don't line up with their respective submissions
- Duplicate code in problem Edit/Create
- ~~The bane of my existence~~ Form validation

I move that we merge now and deal with the leftovers in the (very) near future. It doesn't give me warm fuzzies, but except for the scoreboard it's functional. In the future, PR's probably shouldn't be this big.

BrandonLMorris:
I know you're busy @WilliamHester so I'm just gonna merge this in the next couple of days unless you have an issue with it. It's only like 10-15 lines different than the last time you reviewed it.

WilliamHester:
It looks good! Great work. Yeah, in the future, pull requests will be smaller, but sometimes this stuff just happens.

PR #46, opened 2015-12-22
Scoreboard problems fix

### LINKED ISSUE #40: Scoreboard problems don't line up with submissions
Now that scoreboard uses the problem names, the submissions are not ordered correctly.

This bug was introduced by me. I shame.

WilliamHester:
Make sure everything looks good here. I think it's pretty solid, but I also wasn't able to drive when I wrote this code. Fixes  #40.


auacm/app/static/js/controllers/JudgeController.js
```diff
@@ -29,10 +29,10 @@ app.controller('JudgeController', ['$scope', '$http', '$routeParams', '$window',
             submission.testNum = 0;
             $scope.submitted.push(submission);
         }, function(response) {
-                console.log("error");
+            console.log("error");
```
BrandonLMorris: Can we make this a bit more helpful? At least for admins, it would be nice to know _what_ went wrong. Just printing the response object would be enough.

auacm/app/static/js/controllers/ScoreboardController.js
```diff
@@ -38,26 +37,26 @@ app.controller('ScoreboardController', ['$scope', '$http', '$routeParams',
         var prevSolved = $scope.teams[0].solved;
         var prevTime = $scope.teams[0].solved;
         $scope.teams[0].rank = rank;
-        for (var i = 1; i < $scope.teams.length; i++) {
-            var team = $scope.teams[i];
-            if (team.solved < prevSolved) {
+        for (var j = 1; j < $scope.teams.length; j++) {
```
BrandonLMorris: Bit picky, but is there a reason why this is `j`? Also not sure how I feel about the `team1` name.

auacm/app/modules/scoreboard_manager/views.py
```diff
@@ -70,12 +73,13 @@ def getCompetitionData(cid):
 
     for team in team_users.keys():
```
BrandonLMorris: `.keys()` is unnecessary: dictionaries are directly iterable.

auacm/app/modules/scoreboard_manager/views.py
```diff
@@ -70,12 +73,13 @@ def getCompetitionData(cid):
...
         team_problems = dict()
-        for problem in comp_problems:
+        for name in comp_problems.keys():
```
BrandonLMorris: See above

auacm/app/modules/scoreboard_manager/views.py
```diff
@@ -70,12 +73,13 @@ def getCompetitionData(cid):
...
+            problem = comp_problems[name]
             correct = 0
             incorrect = 0
             pointless = 0
             for s in submissions:
-                if not s.pid == problem or not s.username in team_users[team]:
+                if not s.pid == problem['pid'] or not s.username in team_users[team]:
```
BrandonLMorris: This is picky, but it is more readable and (I think) more pythonic to say `<somthing> not in <list>` rather than `not <something> in <list>`

BrandonLMorris:
Ok, I haven't gotten a chance to run it, but so far it looks good. Just a few things I wanted to point out:

**_Should Fix**_
- [x]  I think you're behind master. Not sure if you were waiting to merge it in.
- [x] `.keys()` isn't necessary to iterate over a dictionary
- [x] Not sure about some of the names in the ScoreboardController

**_If You Feel Like It**_
- The get_competition_data method is a bit complex. I don't know how, but it might could benefit from being broken up some.
- On a similar note to ^, the create_competition_object may fit better in the models.py file as a method on the Competition class.

auacm/app/static/js/controllers/ScoreboardController.js
WilliamHester: The JavaScript linter in Atom was yelling at me that `i` and `team` had already been declared. I think it was a scoping issue. I can change them to be more apt; I'm not sure if I should change them back though.

BrandonLMorris: [Apparently](https://stackoverflow.com/questions/26174520/reusing-i-for-javascript-for-loops) JavaScript is _function-level scoped_ (I think) so the `i` stays in scope after the first `for` loop. Therefore, the second loop is actually _redeclaring_ `i`. However, redeclaring doesn't throw a runtime error, just a linting error. IMO, I'd rather keep it consistent and deal with/suppress the linter errors; my reason is that when I see `j` as a counter variable, it looks like it's a nested loop, which it's not.

**_tl;dr**_ JavaScript is weird.

WilliamHester: Yeah, I actually found that exact SO post. I fixed it by, at the top of the function, declaring 

```
var team, i;
```

I think that's the cleanest way of doing it.

auacm/app/static/js/controllers/ScoreboardController.js
```diff
@@ -107,22 +107,22 @@ app.controller('ScoreboardController', ['$scope', '$http', '$routeParams',
         .then(function(response) {
             $scope.competition = response.data.data.competition;
             $scope.compProblems = response.data.data.compProblems;
-            $scope.problemNames = response.data.data.names;
+            $scope.problemNames = Object.keys(compProblems).sort();
```
BrandonLMorris: This gives an error. Instead of `compProblems` it should be `$scope.compProblems`

auacm/app/static/html/scoreboard.html
```diff
@@ -11,7 +11,7 @@ <h2 class="pull-right" ng-show="active">{{ timeLeft | secondsToDateTime | date:"
           <th>Team</th>
           <th>Solved</th>
           <th>Time</th>
-          <th ng-repeat="name in problemNames">{{ name }}</th>
+          <th ng-repeat="name in problemNames"><a href="/#/problems/{{ problemData[name].shortname }}">{{ name }}</a></th>
```
BrandonLMorris: This link isn't working for me; it just goes back to the problem page. I'm looking into it now

WilliamHester: I think this is related to the above issue.

Edit: I take that back... weird.

BrandonLMorris: Got it (if you haven't already). It should be `compProblems[name].shortname`

PR #49, opened 2015-12-28
Create competition date

WilliamHester:
This does about half of what's required for the pull requests, as it does not allow the creation of teams or registering for a competition as a user. That still needs to be fixed; I'm not sure what the best route is, but I have some ideas for it (hello, table joins).


auacm/app/modules/competition_manager/views.py
```diff
@@ -31,6 +32,82 @@ def get_competitions():
     })
 
 
+@app.route('/api/competitions', methods=['POST'])
+@login_required
+def create_competition():
+    data = request.form
+    if current_user.admin == 0:
+        return serve_error('Only admins can create competitions', 401)
+    if not data['name'] or not data['start_time'] or not data['length'] or\
```
BrandonLMorris: From what I understand, it's more pythonic to [ask for forgiveness rather than permission](http://docs.quantifiedcode.com/python-code-patterns/readability/asking_for_permission_instead_of_forgiveness_when_working_with_files.html). (Yes, the link talks about files specifically, but the principle is the same.) In that sense, it may be better to put this in a try/except block an assume that the form has all the elements. (See problem_manager/views.py create_problem() method for an example)

auacm/app/modules/competition_manager/views.py
```diff
@@ -31,6 +32,82 @@ def get_competitions():
...
+            not data['problems']:
+        return serve_error('You must specify name, startTime,' +
```
BrandonLMorris: The `+` is unnecessary; Python will [automatically concatenate strings](https://stackoverflow.com/questions/18842779/string-concatenation-without-operator-in-python)

auacm/app/modules/competition_manager/views.py
```diff
@@ -31,6 +32,82 @@ def get_competitions():
...
+        name=data['name'],
+        start=int(data['start_time']),
+        stop=(int(data['start_time']) + int(data['length']))
+    )
+    competition.commit_to_session()
+
+    comp_problems = loads(data['problems'])
+    for problem in comp_problems:
+        session.add(CompProblem(
+            label=problem['label'],
```
BrandonLMorris: I guess I should have asked this last night, but this is set up so that the client picks the labels for the problems? In that case, you may want to do a substring to ensure the label isn't to long: something like `label=problem['label'][:2]`

auacm/app/modules/competition_manager/views.py
```diff
@@ -31,6 +32,82 @@ def get_competitions():
...
+    with that competition and add all of the ones in the form body.
+
+    TODO: form validation to make sure that no duplicates are added.
+    """
+    if current_user.admin == 0:
+        # admins only
+        return serve_error('Only admins can modify competitions', 401)
+
+    data = request.form
+    if not data['name'] or not data['start_time'] or not data['length'] or\
```
BrandonLMorris: See above

auacm/app/static/js/controllers/EditCompetitionController.js
```diff
@@ -0,0 +1,223 @@
+app.controller('CreateCompetitionController', ['$scope', '$http', '$location',
+        '$routeParams', function($scope, $http, $location, $routeParams) {
+    $scope.compProblems = [];
+    var competition;
+
+    if ($routeParams.cid !== undefined) {
```
BrandonLMorris, line 6: In my code I checked for undefined variables using `typeof someVar === 'undefined'`. It seems that these two ways do [about the same thing](https://stackoverflow.com/questions/3390396/how-to-check-for-undefined-in-javascript), but I'd like to have one way that we decide and stick to. What do you think?

auacm/app/static/js/controllers/EditCompetitionController.js
```diff
@@ -0,0 +1,223 @@
...
+            $scope.compProblems = [];
+            $location.path('/competitions/' + response.data.data.cid);
+        }, function(response) {
+            $scope.disableForm = false;
+            console.error(response);
+        });
+    };
+}]);
+
+app.directive('dateFormatValidator', [function() {
```
BrandonLMorris, line 134: This is cool, how did you find to use directives for form validation? I used something much different when I was doing user creation.

auacm/app/templates/index.html
```diff
@@ -90,5 +90,6 @@
   <script src="/static/js/controllers/SettingsController.js"></script>
   <script src="/static/js/controllers/ViewProblemController.js"></script>
   <script src="/static/js/controllers/EditProblemController.js"></script>
+  <script src="/static/js/controllers/EditCompetitionController.js"></script>
```
BrandonLMorris, line 93: I didn't see a bunch of duplicate code, but I thought I should bring it up: would it be wise to try to combine the CreateCompetitionController and EditCompetitionController like we did with problems? Again, I'm not advocating for it, I just wanted to throw it out there.

BrandonLMorris:
Cool stuff. I haven't gotten a chance to run it (my terminal is on the fritz, I'm not sure why) but I'll try it soon so we can get this through.

**_Needs to be Addressed**_
- [x] Asking for forgiveness instead of permission in Python (picky)
- [x] Decide how we check undefined variables in JavaScript
- [x] Substring labels (this probably falls under validation, which you explicitly said you're not doing. Sorry)

**_Additional Remarks**_
- Doing form validation with Angular directives is cool, we should look into the (dis)advantages of doing that across the board.
- Cost/benefit of combining competition creation/editing.
- Right now the client picks the labels for problems. To play devil's advocate, why doesn't the API just assign labels based on the order in the problem list?

auacm/app/templates/index.html
WilliamHester, line 93: I think they are combined. I just realized, however, that I forgot to rename the controller in the `app.controller(...` definition.

auacm/app/static/js/controllers/EditCompetitionController.js
WilliamHester, line 134: I actually don't know what made me think to do this. It was probably StackOverflow somewhere, but I'm not sure, to be honest.

auacm/app/static/js/controllers/EditCompetitionController.js
WilliamHester, line 6: I vote for `someVar === undefined`. To me, it's closer to the normal null check that would be done in any sensible language. As someone mentioned in the SO question, it's more concise than `typeof someVar === 'undefined'`. To me, it just makes more sense.

BrandonLMorris, line 6: Sounds good to me. So long as we (nor any library we use) redefines `undefined`, it shouldn't be a problem.

WilliamHester, line 6: I vote that if we find that a library redefines `undefined`, we don't use it, but we can cross that bridge when we come to it.

auacm/app/modules/competition_manager/views.py
WilliamHester: Good catch.

WilliamHester:
Looks like I addressed everything. As for the last 2 additional remarks,
- It looks like the competition creation/editing is combined, unless you're talking about on the Python side.
- The API doesn't assign labels because in the near future, we'll be able to arrange the problems in the order that we want for a competition. If we want to change to a number for order, that's possible as well, but it doesn't make a huge difference to me.

BrandonLMorris:
Ok, I think I see: we're explicitly defining the order using labels rather than implicitly ordering them by the problem list. Correct?

WilliamHester:
Correct.

BrandonLMorris:
In that case, could/should we make labels an optional parameter, and the API can set some arbitrary default if the user fails to provide it? It may not be worth the trouble.

WilliamHester:
I think that, for now, they should be required. If we decide in the future to allow a client not to specify them, then we can handle that then.

BrandonLMorris:
Works for me.

BrandonLMorris:
So setting the time of a competition uses a 24hr clock, right? Can you put a comment or something there so that the user can see that? Or maybe an additional dropdown that specifies AM/PM.

PR #50, opened 2015-12-29
Testing improvements

### LINKED ISSUE #23: Testing
Pretty much everything needs a unit test associated with it. Some integration testing wouldn't hut. TDD all the way, baby.

BrandonLMorris:
Hooray for unit tests! This PR lays the groundwork for testing Flask view methods using a test database. It also includes several unit tests for the problem module (create, get, edit, delete). Modules can now be selected to be tested individually in the `test.py` script using command line arguments.

A few of the setup scripts were modified as well. The `initialize_database.sh` now prompts the user to reset the app `data` file automatically, deleting and reinstating all problem and submission directories. In addition, a `backup_database.sh` script was added that backs up the databases with all the appropriate flags.

For proper testing, a test database was added. As it currently stands, it includes a copy of all the data from the normal `acm` database. All tests are done off the `acm_test` database. `TEST_USERNAME` and `TEST_PASSWORD` configuration fields are defined in `config.py` and are used for logging in/out of the app during tests.

This PR moves us significantly forward to our goal of TDD and assurance that our app works correctly.

Fixes #23 (partially). After this, issues can be created for tests on a per module basis.


auacm/app/database.py
```diff
@@ -3,14 +3,23 @@
...
+from app import app
+
+# Use the test DB if specified, otherwise the default
 Base = automap_base()
 engine = create_engine('mysql+pymysql://acm@localhost/acm?charset=utf8')
 connection = engine.connect()
 session = Session(engine)
+
+# Set up the test database connection
+test_engine = create_engine(
```
WilliamHester: Is the test_engine connected even if we are not testing?

docs/api.md
```diff
@@ -0,0 +1,171 @@
...
+5. Blog Management
+
+---
+
+## Problem Management
+
+### Create a New Problem
+***This method requires being logged in as an administrator***
+
+__URL:__ `/api/problems/``
```
WilliamHester: Accidental extra "`".

docs/api.md
```diff
@@ -0,0 +1,171 @@
+# AUACM API Documentation
+
+The Auburn ACM website uses a RESTful api for managing blog posts, problems,
+submissions, etc. This document is intended to serve as a one stop shop for the
+method arguments and return values.
+
+### Table of Contents
+
+1. Problem Management
```
WilliamHester: Just for convenience, in the table of contents, you can do something like `[Problem Management](#problem-management)`, which will link to the corresponding section of the doc.

setup/initialize_database.sh
```diff
@@ -1,9 +1,36 @@
...
+fi
+
+# Create the data directory if it doesn't exist
+if [ ! -e ../auacm/app/data ]
+then
+  echo "Setting up submissions and problems data."
+  mkdir ../auacm/app/data
+  cp data.zip ../auacm/app/data.zip
+  cd ../auacm/app
+  unzip data.zip
```
WilliamHester: I know we didn't do this in the past, but it's kind of annoying to see all of the output for each file that is unzipped. Can we instead do `unzip data.zip > /dev/null`?

setup/initialize_database.sh
```diff
@@ -1,9 +1,36 @@
...
+if [ -e ../auacm/app/data ]
+then
+  echo "Do you want to reset the data directory? (y/n)"
+  read purge
+fi
+if [ "$purge"="y" ]
+then
+  echo "Purging and resetting data directory..."
+  rm -rf ../auacm/app/data
+  rmdir ../auacm/app/data
```
WilliamHester: I think this is unnecessary. I get an error from this line when I run the script.

setup/initialize_database.sh
```diff
@@ -1,9 +1,36 @@
...
+echo "Setting up mysql database..."
 sudo mysql -uroot < acm.sql
-echo "Setting up submissions and problems data."
-mkdir ../auacm/app/data
-cp data.zip ../auacm/app/data/data.zip
-cd ../auacm/app/data
-unzip data.zip
-rm data.zip
+echo "Setting up test database..."
+echo "DROP DATABASE IF EXISTS acm_test; CREATE DATABASE acm_test; USE acm_test;\n" | mysql -uroot < acm.sql
```
WilliamHester: This isn't working at all for me. The acm.sql file actually has the line `USE acm;`, so we should probably find a better way of setting it up.

auacm/app/database.py
BrandonLMorris: Yes. Not ideal, I'll admit it, but it's the best I could come up with. This get called from app/**init** so there's no way to change the config programmatically before the connection is established. This is also in part a result of the global variable madness. It can be changed, I think.

setup/initialize_database.sh
BrandonLMorris: I think it's supposed to read from `acm_test.sql`. I'll look into it.

BrandonLMorris:
Ok, that should address everything you've mentioned thus far. It should also actually work now.
- Test database only gets connected if actual testing is being done.
- DB scripts ensure existence of acm_test.sql and properly create the test database
- Docs removed typo and added relative links

docs/api.md
```diff
@@ -0,0 +1,170 @@
...
+  * __date added__
+  * __solved by user__
+  * __problem url__
+
+of all the problems.
+
+## Edit an Existing Problem
+***This method requires being signed in as an administrator***
+
+__URL:__ `/api/problems/{identifier}/edit`
```
WilliamHester: This should not contain `/edit`. Sorry for forgetting about this earlier.

WilliamHester:
Looks good!

PR #54, opened 2016-01-02
Competition teams

### LINKED ISSUE #24: Create competitions
Competitions should be able to be created by an admin inside the web app, rather than interacting with the DB directly.

### LINKED ISSUE #25: Edit competitions
Competitions should be editable after they have been created. They should probably not be editable once the competition has passed. Things that should be editable include: Users in contest, problems in contest, start/end times.

WilliamHester:
Allows creation/modification of teams for competitions. Fixes #24 and fixes #25 when combined with #49.  Merge #49 first, as this is based off of that.


auacm/app/modules/competition_manager/views.py
```diff
@@ -181,3 +202,129 @@ def get_competition_data(cid):
...
+        return serve_error('Competition does not exist', response_code=404)
+
+    if current_user.admin == 1 and 'users' in request.data:
+        registrants = loads(request.data['users'])
+    else:
+        registrants = list()
+        registrants.append(current_user.username)
+
+    for user in registrants:
+        if session.query(CompUser).filter(CompUser.cid == cid,\
```
BrandonLMorris: I'm not going to make you change it (I know we've done it elsewhere), but Google style [forbids backslash continuation.](https://google.github.io/styleguide/pyguide.html?showone=Line_length#Line_length)

I'm not sure it's 100% necessary in the places you used it, particularly around [parens, brackets, and braces](http://docs.python.org/reference/lexical_analysis.html#implicit-line-joining).

auacm/app/modules/competition_manager/views.py
```diff
@@ -181,3 +202,129 @@ def get_competition_data(cid):
...
+
+@app.route('/api/competitions/<int:cid>/register', methods=['POST'])
+@login_required
+def register_for_competition(cid):
+    """ Called when a user wants to register for a competition.
+
+    All the user has to do is submit a post to this url with no form data.
+    From their logged-in status, we'll go ahead and add them to the competiton
+    as an individual (team name is default their display name).
+    """
```
BrandonLMorris, line 226: Can this say something about how it works with an admin? I'd like to have it in writing.

auacm/app/modules/competition_manager/views.py
```diff
@@ -181,3 +202,129 @@ def get_competition_data(cid):
...
+    if session.query(Competition).filter(Competition.cid == cid).first() \
+            is None:
+        return serve_error('Competition does not exist', response_code=404)
+
+    if False and current_user.admin == 1 and 'users' in request.data:
```
BrandonLMorris: `if False`? That doesn't seem right...

auacm/app/modules/competition_manager/views.py
```diff
@@ -181,3 +202,129 @@ def get_competition_data(cid):
...
+
+@app.route('/api/competitions/<int:cid>/unregister', methods=['POST'])
+@login_required
+def unregister_for_competition(cid):
+    """ Called when a user wants to register for a competition.
+
+    All the user has to do is submit a post to this url with no form data.
+    From their logged-in status, we'll go ahead and add them to the competiton
+    as an individual (team name is default their display name).
+    """
```
BrandonLMorris, line 278: See above about admin functionality

auacm/app/modules/competition_manager/views.py
```diff
@@ -181,3 +202,129 @@ def get_competition_data(cid):
...
+            session.add(CompUser(
+                cid=cid,
+                username=user,
+                team=team
+            ))
+
+    session.flush()
+    session.commit()
+
+    return serve_response({})
```
BrandonLMorris, line 363: Should this return the new team or something?

WilliamHester, line 363: If we returned the teams, we'd just be returning the JSON data that the user just specified in the PUT. It would be kind of redundant. All we need to know here is that the request was successful.

auacm/app/modules/competition_manager/views.py
```diff
@@ -18,13 +24,26 @@ def get_competitions():
...
+    registered = set()
+    for row in registered_rows:
+        registered.add(row.cid)
+
     for competition in session.query(Competition).all():
         if competition.stop < current_time:
-            past.append(competition.to_dict())
+            past.append(competition.to_dict(
+                user_registered=competition.cid in registered
+            ))
```
BrandonLMorris: Just want to note that [Google style](https://google.github.io/styleguide/pyguide.html?showone=Indentation#Indentation) has closing parens for function calls on the same line

auacm/app/modules/competition_manager/views.py
```diff
@@ -181,3 +202,149 @@ def get_competition_data(cid):
...
+    """
+    if session.query(Competition).filter(Competition.cid == cid).first() \
+            is None:
+        return serve_error('Competition does not exist', response_code=404)
+
+    if current_user.admin == 1 and 'users' in request.data:
+        registrants = loads(request.data['users'])
+    else:
+        registrants = list()
+        registrants.append(current_user.username)
```
BrandonLMorris: These two lines can be combined to `registrants = [current_user.username]`

auacm/app/modules/competition_manager/views.py
```diff
@@ -181,3 +202,149 @@ def get_competition_data(cid):
...
+    session.flush()
+    session.commit()
+
+    return serve_response({})
+
+
+@app.route('/api/competitions/<int:cid>/unregister', methods=['POST'])
+@login_required
+def unregister_for_competition(cid):
+    """ Called when a user wants to register for a competition.
```
BrandonLMorris: When the user wants to **unregister**

auacm/app/modules/competition_manager/views.py
```diff
@@ -181,3 +202,149 @@ def get_competition_data(cid):
...
+    """
+    if session.query(Competition).filter(Competition.cid == cid).first() \
+            is None:
+        return serve_error('Competition does not exist', response_code=404)
+
+    if current_user.admin == 1 and 'users' in request.data:
+        registrants = loads(request.data['users'])
+    else:
+        registrants = list()
+        registrants.append(current_user.username)
```
BrandonLMorris: Can be combined (see above)

auacm/app/modules/competition_manager/views.py
```diff
@@ -181,3 +202,149 @@ def get_competition_data(cid):
...
+
+    for user in registrants:
+        (session.query(CompUser)
+                .filter(CompUser.username == user, CompUser.cid == cid)
+                .delete())
```
BrandonLMorris, line 295: I like the parens over backslash. I'm pretty sure that follows Google style, unlike some of our legacy code. We should move forward with this and (at some point) change the old ones.

auacm/app/modules/competition_manager/views.py
```diff
@@ -181,3 +202,149 @@ def get_competition_data(cid):
...
+
+
+@app.route('/api/competitions/<int:cid>/teams', methods=['PUT'])
+@login_required
+def put_competition_teams(cid):
+    """ Update the teams for a competition
+
+    If a user is an admin, they can update the competition's users, doing a PUT.
+    This will take the JSON data in the 'teams' part of the request form and
+    store it to the database. Any teams or users not included int the JSON data
```
BrandonLMorris: "...not included **int** the JSON..."

auacm/app/modules/competition_manager/views.py
```diff
@@ -181,3 +202,149 @@ def get_competition_data(cid):
...
+    will not be a part of the competition and will have to re-register; however
+    it should not be used for the solely purpose of de-registering participants.
+    """
+    if current_user.admin == 0:
+        return serve_error('You must be an admin to update teams',
+                response_code=401)
+
+    try:
+        teams = loads(request.form['teams'])
+    except KeyError as err:
```
BrandonLMorris, line 342: It may be a good idea to catch a ValueError if the JSON isn't properly formed. Kinda falls under form validation

BrandonLMorris:
What does the `closed` field do? Does it just specify if a given competition is active? Is it necessary, or can it just be computed from start/end times?

The code to a competition needs to be updated to include it, if it's going to remain. Right now trying to create a comp in the app throws a SQLAlchemy IntegrityError saying that "'closed' cannot be null"

WilliamHester:
The `closed` field is for competitions that can only have users registered by admins. It's really for a feature that hasn't been implemented yet but will definitely be implemented soon. I'm not sure why that error is being thrown (probably because I forgot to assign a value to it in the competition creation).

BrandonLMorris:
You can create a default value for it in models.py, but you have to be a little careful about. You can see what I did in problem_manager/models.py

auacm/app/modules/competition_manager/views.py
WilliamHester, line 342: Done. I went ahead and added this check everywhere in this file where we used `loads()`.

auacm/app/static/js/controllers/TeamController.js
```diff
@@ -0,0 +1,101 @@
...
+            teams[team] = [];
+            for (i = 0; i < $scope.teams[team].length; i++) {
+                teams[team].push($scope.teams[team][i].username);
+            }
+        }
+        for (i = 0; i < $scope.individuals.length; i++) {
+            teams[$scope.individuals[i].display] =
+                    [$scope.individuals[i].username];
+        }
+        console.log(teams);
```
BrandonLMorris, line 67: This looks like it might have been left over from debug? Just wanted to point it out.

BrandonLMorris:
Looks good. TIME TO BETA!

PR #56, opened 2016-01-22
Fix crashes from string encoding and database column limit

WilliamHester:
Okay, so I was trying to submit a problem and I got [this error](https://gist.github.com/WilliamHester/571d3882c7dfc24ced58). Eventually, I boiled it down to our casting of `request.form['cases']` to a string type. After removing the cast, I ran into a separate yet related issue where the column only allowed for sample cases that were 255 characters or less, which was not enough for some of the old-style problems, since, technically, there's only one test case.

### DIFF
--- a/auacm/app/modules/problem_manager/views.py
+++ b/auacm/app/modules/problem_manager/views.py
@@ -129,7 +129,7 @@ def create_problem():
     try:
         # Convert the JSON to python array of dictionaries
         cases = request.form['cases']
-        cases = loads(str(cases))
+        cases = loads(cases)
         for case in cases:
             if 'input' not in case or 'output' not in case:
                 return serve_error(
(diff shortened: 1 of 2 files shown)


BrandonLMorris:
Is this a result of PDF's weird encoding of whitespace? Were you trying to directly copy/paste into the text box?

I looked it up, the issue is a non-breaking space (unicode 160). The real issue is that it's Python 2, where not all strings are Unicode. The error wouldn't have happened in Python 3. `loads()` accepts a `sting` or `unicode` object (see [docs](https://docs.python.org/2/library/json.html#json.loads)) so that's why removing the cast works.

That being said, I agree about upping the 255 character limit. I thought it might be appropriate, I should have seen that it wasn't. 

> _"Premature optimization is the root of all evil"_

PR #57, opened 2016-01-25
Fix annoying login bug for now

WilliamHester:


auacm/app/static/js/controllers/LoginController.js
```diff
@@ -20,9 +20,9 @@ app.controller('LoginController', ['$scope', '$http', '$window',
                     },
                     data: request
                 }).then(function(response) {
-                        $window.location.href = 'http://' + $window.location.host + '/#/';
+                        window.location.href = 'http://' + $window.location.host + '/#/';
```
BrandonLMorris: Is this supposed to be like this?

PR #59, opened 2016-01-31
Fix competitions time-related issues

### LINKED ISSUE #55: Scoreboard doesn't automatically start
When awaiting the beginning of a competition, the scoreboard will remain inactive even after the start time has passed. A refresh or reload is necessary for it to begin ticking. Not only is it unintuitive, it invites users to hammer the site with refreshes before a competition begins.

WilliamHester:
I didn't mean to do two things at once here, but I did. I fixed the issue of competitions displaying the incorrect start time and also fixed the issue if competitions not starting the timer when the competition started. While I was there, I went ahead and cleaned up the code that displays the timer and added variables to denote different states of the competition, `active`, as in "ongoing", and `ended`. This enabled me to hide the problems before the competition started on the scoreboard page, showing them once the clock starts ticking.

Fixes #55

### DIFF
--- a/auacm/app/static/html/scoreboard.html
+++ b/auacm/app/static/html/scoreboard.html
@@ -1,7 +1,7 @@
 <div>
   <div class="row">
     <h1 class="pull-left">{{ competition.name }}</h1>
-    <div ng-if="isAdmin && (active || !over)" class="dropdown pull-left margin-top margin-left">
+    <div ng-if="isAdmin && !ended" class="dropdown pull-left margin-top margin-left">
       <button class="btn btn-default dropdown-toggle" type="button"
           id="dropdownMenu1" data-toggle="dropdown" aria-haspopup="true"
           aria-expanded="true">
--- a/auacm/app/static/html/scoreboard.html
+++ b/auacm/app/static/html/scoreboard.html
@@ -1,7 +1,7 @@
 <div>
   <div class="row">
     <h1 class="pull-left">{{ competition.name }}</h1>
-    <div ng-if="isAdmin && active" class="dropdown pull-left margin-top margin-left">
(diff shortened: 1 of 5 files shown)


auacm/app/static/js/controllers/EditCompetitionController.js
```diff
@@ -10,7 +10,9 @@ app.controller('EditCompetitionController', ['$scope', '$http', '$location',
             .then(function(response) {
                 competition = response.data.data.competition;
 
-                var date = new Date(competition.startTime * 1000);
+                var offset = Date.now().getTimezoneOffset() / 1000;
+                console.log(offset);
```
BrandonLMorris: Is this leftover or intentional?

auacm/app/static/html/scoreboard.html
```diff
@@ -1,7 +1,7 @@
 <div>
   <div class="row">
     <h1 class="pull-left">{{ competition.name }}</h1>
-    <div ng-if="isAdmin && active" class="dropdown pull-left margin-top margin-left">
+    <div ng-if="isAdmin && (active || !over)" class="dropdown pull-left margin-top margin-left">
```
BrandonLMorris: is this supposed to be `!ended`? What is the difference between `over` and `ended`?

auacm/app/static/js/controllers/EditCompetitionController.js
WilliamHester: Oops. You're right. This is leftover.

auacm/app/static/html/scoreboard.html
WilliamHester: `over` became `ended`, and I forgot to change it.

BrandonLMorris:
Kewl sauce.

PR #60, opened 2016-01-31
Anonymous user

WilliamHester:
This one was fun. I finally went ahead and made our fabled `@admin_required` decorator, which now requires that the user that is accessing the endpoint is logged in and an admin. On a much larger note, this code allows a user to be logged out and still access the site. Any `POST` functions are disabled, other than `/api/login` (they keep the `@login_required` decorator). Any buttons that would submit anything are also disabled, and the judge page is hidden from the navbar. This still needs a bit of polish, as there's an annoying bug when you try to log in using the new dropdown, but it's pretty close.


auacm/app/util.py
```diff
@@ -39,3 +41,13 @@ def serve_response(response, response_code=200):
 def serve_error(error, response_code):
     '''Serve an error in response to a request.'''
     return jsonify({'status': response_code, 'error': error}), response_code
+
+def admin_required(function):
+    @wraps(function)
+    def wrap(*args, **kwargs):
+        if current_user.is_anonymous or current_user.admin == 0:
+            return serve_error('You need to be an admin to do that.',
```
BrandonLMorris, line 49: I wouldn't mind this being a little bit more specific. I'm not sure, but maybe the decorator could take an argument with a string describing the action, so `@admin_required` would be `@admin_required('create a problem')`. Probably not worth fixing in this PR

BrandonLMorris:
Looks good, let me just give it a test run before I merge it. By the way, did you happen to run the unit tests?

BrandonLMorris:
Awesome, looks great! I'm glad we finally have this.

I will mention that an anonymous user can still get to the judge page by changing the URL to `#/judge`, but they can't submit or upload a problem.

PR #61, opened 2016-02-07
Judge bugs

WilliamHester:
Fix one bug here and added Python3 submission support. The bug was that over a network (with a delay), sorting the problems in the judge controller would break because the problems may not have existed yet.


auacm/app/modules/submission_manager/judge.py
```diff
@@ -77,10 +80,10 @@ def compile_submission(submission, uploaded_file):
     filename = uploaded_file.filename
     name, ext = filename.rsplit('.', 1)
```
BrandonLMorris, line 81: Should `ext` here also be `_` like in line 225? Doesn't change anything functionally, just and unused variable.

BrandonLMorris:
Cool sauce, looks great.

PR #64, opened 2016-02-09
Fix python 3 support

### LINKED ISSUE #20: Socket should be Python 3 compatible
Right now, the library used for Python web sockets is only for Python 2.7. This is the only thing holding us back from being completely Python 3.

WilliamHester:
By switching to a Tornado server, I was able to provide support for WebSockets natively.

Also fixes #20

### DIFF
--- a/auacm/app/modules/competition_manager/views.py
+++ b/auacm/app/modules/competition_manager/views.py
@@ -1,8 +1,9 @@
 from flask import request
 from flask.ext.login import current_user, login_required
-from app import app, socketio
+from app import app
 from app.database import session
 from app.util import serve_response, serve_error, admin_required
+from app.modules.flasknado.flasknado import Flasknado
 from .models import Competition, CompProblem, CompUser
 from app.modules.submission_manager.models import Submission
 from app.modules.problem_manager.models import Problem
@@ -10,11 +11,6 @@ from app.modules.user_manager.models import User
 from sqlalchemy import asc
 from time import time
 from json import loads
-
-
-@socketio.on('connect', namespace='/register')
-def on_connection():
-    pass
 
 
 @app.route('/api/competitions')
@@ -156,7 +152,7 @@ def get_competition_data(cid):
 
     scoreboard = list()
 
-    team_users = dict()
+    _users = dict()
     for user in comp_users:
         if not user.team in team_users:
             team_users[user.team] = list()
@@ -222,9 +218,9 @@ def register_for_competition(cid):
             is None:
         return serve_error('Competition does not exist', response_code=404)
 
-    if current_user.admin == 1 and 'users' in request.data:
+    if current_user.admin == 1 and 'users' in request.form:
         try:
-            registrants = loads(request.data['users'])
+            registrants = loads(request.form['users'])
         except ValueError:
             return serve_error('JSON data for \'users\' not properly formatted',
                     response_code=400)
@@ -244,14 +240,13 @@ def register_for_competition(cid):
             username=user.username,
             team=user.display
         ))
-        socketio.emit('new_user', {
+        Flasknado.emit('new_user', {
             'cid': cid,
             'user': {
                 'display': user.display,
                 'username': user.username
             }
-        },
-        namespace='/register')
+        })
     session.flush()
     session.commit()
 
@@ -274,9 +269,9 @@ def unregister_for_competition(cid):
             is None:
         return serve_error('Competition does not exist', response_code=404)
 
-    if current_user.admin == 1 and 'users' in request.data:
+    if current_user.admin == 1 and 'users' in request.form:
         try:
-            registrants = loads(request.data['users'])
+            registrants = loads(request.form['users'])
         except ValueError:
             return serve_error('JSON data for \'users\' not properly formatted',
                     response_code=400)
--- a/auacm/app/modules/competition_manager/views.py
+++ b/auacm/app/modules/competition_manager/views.py
@@ -152,7 +152,7 @@ def get_competition_data(cid):
 
     scoreboard = list()
 
-    _users = dict()
+    team_users = dict()
     for user in comp_users:
         if not user.team in team_users:
             team_users[user.team] = list()
--- a/auacm/app/modules/flasknado/flasknado.py
+++ b/auacm/app/modules/flasknado/flasknado.py
@@ -32,7 +32,6 @@ class Flasknado(WebSocketHandler):
             })
 
     def open(self):
-        print('opening connection')
         Flasknado.clients.append(self)
 
     def on_close(self):
--- a/auacm/app/modules/flasknado/flasknado.py
+++ b/auacm/app/modules/flasknado/flasknado.py
@@ -1,0 +1,45 @@
+from tornado.websocket import WebSocketHandler
+
+def make_flasknado():
+    """ Creates a new class instance of Flasknado
+
+    In a sense, this is kind of a hack. The problem is that the way Tornado
+    works, it requires sending a WebSocketHandler class rather than an instance
+    of that class. This means that in order to have separate instances, we must
+    dynamically create different classes at runtime. This function simply
+    returns a Flasknado class with an empty set of clients.
+    """
+
+    values = dict(Flasknado.__dict__)
+    values['clients'] = []  # give the new instance a new set of clients
+    return type('Flasknado', Flasknado.__bases__, values)
+
+class Flasknado(WebSocketHandler):
+    """ A class for handling sockets
+
+    This class acts as a singleton that contains all of the websocket clients
+    that are currently connected to the server.
+    """
+
+    clients = []  # initialize the list of clients to an empty list
+
+    @classmethod
+    def emit(self, event_type, data):
+        for client in self.clients:
+            client.write_message({
+                'eventType': event_type,
+                'data': data
+            })
+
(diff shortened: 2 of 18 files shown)


auacm/requirements.txt
```diff
@@ -5,13 +5,14 @@ Flask-Bcrypt==0.7.1
 Flask-Login==0.3.0
 Flask-SocketIO==1.2
```
BrandonLMorris: This is no longer necessary...right?

auacm/app/modules/competition_manager/views.py
```diff
@@ -222,9 +218,9 @@ def register_for_competition(cid):
             is None:
         return serve_error('Competition does not exist', response_code=404)
 
-    if current_user.admin == 1 and 'users' in request.data:
+    if current_user.admin == 1 and 'users' in request.form:
```
BrandonLMorris, line 221: So going from `data` to `form` is because of the tornado server?

auacm/app/modules/flasknado/flasknado.py
```diff
@@ -0,0 +1,45 @@
...
+    return type('Flasknado', Flasknado.__bases__, values)
+
+class Flasknado(WebSocketHandler):
+    """ A class for handling sockets
+
+    This class acts as a singleton that contains all of the websocket clients
+    that are currently connected to the server.
+    """
+
+    clients = []  # initialize the list of clients to an empty list
```
BrandonLMorris, line 24: Um, not exactly sure how this is being used, but it probably needs to go inside the `emit()` method. The problem is that lists are mutable in Python, so even though they can be accidentally shared across multiple objects if they're defined at the class level.

[Here](https://gist.github.com/BrandonLMorris/12c202a414b70c88a6be) is a quick example I made to demonstrate the issue. I'm not sure if it completely applies since (to my knowledge) we're only using this class for it's static method and not actually creating instances, but I thought you should be aware.

BrandonLMorris, line 24: Oh wait, I think that's the desired effect. My bad. That's cool. It's early.

auacm/app/modules/flasknado/flasknado.py
```diff
@@ -0,0 +1,45 @@
...
+    @classmethod
+    def emit(self, event_type, data):
+        for client in self.clients:
+            client.write_message({
+                'eventType': event_type,
+                'data': data
+            })
+
+    def open(self):
+        print('opening connection')
```
BrandonLMorris: Debug? I'm ok with a log statement, but maybe not standard out.

auacm/app/modules/submission_manager/models.py
```diff
@@ -51,17 +51,15 @@ def emit_status(self, status, test_num):
         :param status: the status of the submission
         :return: None
         '''
-        socketio.emit(
-            'status',
-            {
+        print(status)
```
BrandonLMorris: Looks like another debug spot

auacm/app/static/js/controllers/JudgeController.js
```diff
@@ -38,12 +38,9 @@ app.controller('JudgeController', ['$scope', '$http', '$routeParams', '$window',
...
     };
 
-    var socket = io.connect('http://' + $window.location.host + '/judge');
-    socket.on('connect', function() {
-        console.log('connected');
-    });
-
+    var socket = new Socket('ws://' + $window.location.host + '/websocket');
     socket.on('status', function(event) {
+        console.log(event);
```
BrandonLMorris: I don't really have a problem with this, just wanted to make sure you saw it.

auacm/app/static/js/controllers/TeamController.js
```diff
@@ -80,8 +80,7 @@ app.controller('TeamController', ['$scope', '$http', '$location', '$window',
         });
     };
 
-    var socket = io.connect('http://' + $window.location.host + '/register');
-    socket.on('connect', function() {});
+    var socket = new Socket('ws://' + $window.location.host + '/websocket');
```
BrandonLMorris: Since we are, at the moment, a single page application, is there really a need to open connections multiple times? Before we had separate domains, but that may not be necessary and that doesn't seem to be what were doing anymore. Perhaps one `appSocket` defined in MainController that can be accessed by all the subcontrollers. I don't know if that would cause extra overhead or something.

auacm/app/static/js/sockets.js
```diff
@@ -0,0 +1,14 @@
...
+    this.on = function(event, func) {
+        parent.events[event] = func;
+    };
+    this.ws.onmessage = function(event) {
+        var data = JSON.parse(event.data);
+        if (data.eventType in parent.events) {
+            parent.events[data.eventType](data.data);
+        }
+    };
+}
```
BrandonLMorris, line 14: Is this likely to be expanded in the future? If not, I don't see the real need to put it in a separate file and create the overhead of _another_ request to the server.

auacm/requirements.txt
```diff
@@ -5,13 +5,14 @@ Flask-Bcrypt==0.7.1
...
 Flask-SQLAlchemy==2.0
-eventlet==0.18.1
 itsdangerous==0.24
 Jinja2==2.8
+lazy-object-proxy==1.2.1
```
BrandonLMorris: Is this for Tornado?

auacm/app/modules/submission_manager/judge.py
```diff
@@ -65,9 +62,6 @@ def evaluate(submission, uploaded_file, time_limit=10):
     :param uploaded_file: the uploaded file
     :return: the status of the submission (one of the status constants above)
     """
-    directory = directory_for_submission(submission)
-    os.mkdir(directory)
-    uploaded_file.save(os.path.join(directory, uploaded_file.filename))
```
BrandonLMorris, line 70: Removing these breaks the judge tests. I know it's necessary since you can't save the files in a separate thread, but we need to figure that out.

BrandonLMorris, line 70: I think. Not 100% sure. Looking into it. Judge tests are definitely broken though.

BrandonLMorris:
So judge tests are broken in multiple ways. Most are probably not due to this specific PR, but at least some are, and I'd like to address them before a merge.

A lot of the (initial) issues are from certain directories not existing. This is because the submission tests use a separate directory location, and not all of the subdirectories are always created, leading to some issues. This can be fixed with a couple of lines relatively easy:
- Line 185 in `judge.py` needs to be changed from `os.mkdir(output_path)` to `os.makedirs(output_path)`
  - (The `makedirs()` method will create any intermediate subdirectories if they don't exists, whereas `mkdir()` will just error out.)
- Similarly, right after line 73 in `judge.py`, a check needs to be done that the `directory` exists, and if it doesn't, create it. (Creation in normal useage is handled by the route method, but unit testing circumvents that). Add the two lines:

``` python
if (not os.path.exists(directory)):
  os.makedirs(directory)
```

---

After those changes, I'm still getting two kinds of **failures** (they run, just don't pass): all C programs give a `COMPILATION_ERROR` and I'm not sure why (C programs have worked in the past, but it may be a problem with my system). Additionally, a few Python tests cases give runtime errors. My suspicion is that it has something to do with the Python 3 additions we've been doing lately.

BrandonLMorris:
Ok that didn't quite fix what I wanted, and it's not that great of a solution. I may branch off of this and try and fix the judge tests myself.

auacm/requirements.txt
WilliamHester: Right. I forgot to remove it.

auacm/app/modules/competition_manager/views.py
WilliamHester, line 221: I think it's actually because of the python3 switch, not Tornado. I noticed it was broken after I finished the conversion, but it seemed to be because of python3.

auacm/app/modules/flasknado/flasknado.py
WilliamHester: Oops. Yeah, I forgot to remove that.

auacm/app/modules/submission_manager/models.py
WilliamHester: Same

auacm/app/static/js/controllers/JudgeController.js
WilliamHester: Should probably remove this.

auacm/app/static/js/controllers/TeamController.js
WilliamHester: That's not a bad idea. I can't think of any negative consequences. I'll do that.

auacm/app/static/js/sockets.js
WilliamHester, line 14: Yeah, if we do any sort of client communication to the server via sockets, then this will need to be expanded upon. I did just notice, however that I shouldn't be setting some variables to `this.*`. They should be function scope only for encapsulation.

auacm/requirements.txt
WilliamHester: I... I don't know where that came from. Probably Tornado, but I'll check it out.

auacm/app/static/js/sockets.js
BrandonLMorris, line 14: Does that need to be resolved? Tbh I'm not sure what you mean.

PR #65, opened 2016-02-13
Fixes for judge tests

BrandonLMorris:
As part of the upgrade to Python 3, the code to save files
inside the judge module was moved to the main thread. The
unit tests, however, broke because they relied on this
functionality. This commit puts the file saving in the tests
themselves, fixing the tests.


BrandonLMorris:
Just as a side note for posterity's sake, I was having trouble running the go tests, but only when I executed the tests from IntelliJ. I'm not sure how the path gets resolved in `subprocess.call()` and how that changes when run through IntelliJ, but apparently it does not include the /usr/local/bin folder, where go is installed from Homebrew. Creating a symlink the /usr/bin/go solved it.

PR #68, opened 2016-02-18
Scoreboard Timing

BrandonLMorris:
Adds an offset to the scoreboard of the difference between the server's system time and the client's system time. This should sync everybody up, so that no matter the time on the client, the competition will open at the same moment (give or take a few milliseconds).

### DIFF
--- a/auacm/app/modules/competition_manager/views.py
+++ b/auacm/app/modules/competition_manager/views.py
@@ -191,6 +191,9 @@ def get_competition_data(cid):
         team_row['problemData'] = team_problems
         scoreboard.append(team_row)
 
+    # Send the system time to the client
+    Flasknado.emit('system_time', {'milliseconds': int(time() * 1000)})
+
     return serve_response({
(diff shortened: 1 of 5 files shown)


auacm/app/modules/competition_manager/views.py
```diff
@@ -191,6 +191,9 @@ def get_competition_data(cid):
         team_row['problemData'] = team_problems
         scoreboard.append(team_row)
 
+    # Send the system time to the client
+    Flasknado.emit('system_time', {'milliseconds': int(time() * 1000)})
```
WilliamHester: Due to the nature of Flasknado being a class for _all_ clients, every time someone requests a competition from the server, everyone's offset is going to be recalibrated. It might make more sense for the client to request the time then be sent it as a response (still via websocket). It would require a slight modification of `socket.js` but nothing drastic.

Is there any reason you think we should be sending the time to all clients every time it is requested? If so, we can keep it like this.

PR #70, opened 2016-02-21
Problem timing

WilliamHester:
There were a lot of changes made here, but here are the main ones:
- The Judge is now a class. This means that the properties of it are
  simply set when it is created, and calling Judge.run() will start the
  execution of the class.
- Similarly, calling Judge.run_threaded() will start the execution on a
  new thread.
- The Flasknado code has been removed from the Submission model for
  separations of concerns. It doesn't make sense to be doing network
  operations from models.
- Due to run_threaded() already calling a new thread, the JudgeThread
  class has been removed, as it only added unnecessary complexity to the
  code. The logic for terminating a subprocess is entirely handled by the
  while loop in judge_process().

### DIFF
--- a/auacm/app/modules/submission_manager/judge.py
+++ b/auacm/app/modules/submission_manager/judge.py
@@ -3,6 +3,7 @@ import os.path
 import shlex
 import subprocess
 import threading
+import time
 
 from app import app
 from app.modules.flasknado.flasknado import Flasknado
@@ -72,28 +73,34 @@ class Judge:
 
 
     def run_threaded(self):
-        """Runs the Judgement on a new thread"""
+        """Runs the Judgement on a new thread
+
+        :return: the thread that was created for the execution
+        """
         thread = threading.Thread(target=self.run)
         thread.daemon = False
         thread.start()
+        return thread
 
 
     def run(self):
         """Attempts to compile (if necessary) then execute a given file.
 
-        :return: the status of the submission (one of the status constants above)
+        :return: one of the status constants representing the success
         """
-        status = self.compile_submission()
-        self.update_status(status, -1)
+        status = self._compile_submission()
+        self._update_status(status, -1)
 
         if status == COMPILATION_SUCCESS:
-            status, test_number = self.execute_submission()
-            self.update_status(status, test_number)
+            status, test_number, max_time = self._execute_submission()
+            self._update_status(status, test_number)
+        else:
+            max_time = -1
 
-        return status
+        return status, max_time
 
 
-    def update_status(self, status, test_number):
+    def _update_status(self, status, test_number):
         """Updates the status of the submission and notifies the clients that
         the submission has a new status.
         """
@@ -108,7 +115,7 @@ class Judge:
         })
 
 
-    def compile_submission(self):
+    def _compile_submission(self):
         """Compile the submission if it needs compilation. A programming
         language that does not need compilation will return COMPILATION_SUCCESS.
         """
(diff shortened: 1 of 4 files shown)


auacm/app/modules/submission_manager/judge.py
```diff
@@ -33,6 +34,9 @@
     'cpp': 1,
     'go': 1
 }
+DB_STATUS = ['', 'compile', 'start', 'runtime', 'timeout', 'wrong', 'good']
+EVENT_STATUS = ['', 'compile', 'running', 'runtime', 'timeout', 'incorrect',
+                'correct']
```
BrandonLMorris, line 39: Is there any way we could combine these?

WilliamHester, line 39: Yeah, we should be able to do that. I'm going to prefer no changes on the DB though, so we'll have to make a few changes on the front-end. It'll be a lot nicer to see things capitalized anyway.

auacm/app/modules/submission_manager/judge.py
```diff
@@ -42,182 +46,169 @@
...
+                    return TIMELIMIT_EXCEEDED, test_number
+
+                if process.poll() != 0:
+                    return RUNTIME_ERROR, test_number
+
+                result_path = os.path.join(self.directory_for_problem, 'out')
+
+                # The execution is completed.  Check its correctness.
+                with open(os.path.join(output_path, out_file)) as correct_result, \
+                     open(os.path.join(result_path, out_file)) as submission_result:
```
BrandonLMorris: Line continuation with [parentheses](https://google.github.io/styleguide/pyguide.html?showone=Line_length#Line_length) is prefered over backslash.

auacm/app/modules/submission_manager/views.py
```diff
@@ -6,7 +6,7 @@
 from app import app
 from app.util import serve_response, serve_error
 from app.modules.submission_manager import models
-from app.modules.submission_manager import judge
+from app.modules.submission_manager.judge import Judge
```
BrandonLMorris: This causes an error on line 37 since the import has changed.

auacm/app/modules/submission_manager/judge.py
```diff
@@ -42,182 +46,169 @@
...
+        self.directory_for_problem = os.path.join(
+            app.config['DATA_FOLDER'], 'problems', str(self.submission.pid))
+        self.time_limit = time_limit
 
-def directory_for_problem(submission):
-    return os.path.join(app.config['DATA_FOLDER'], 'problems', str(submission.pid))
 
+    def run_threaded(self):
+        """Runs the Judgement on a new thread"""
+        thread = threading.Thread(target=self.run, args=(self))
```
BrandonLMorris: The `args=(self)` is unnecessary and actually breaks. Didn't catch it in the unit tests because `run_threaded()` never gets used (maybe it should get added).

BrandonLMorris:
Looks good, I'm really happy with the way the judge looks now. Just a few quick fixes. Also, I think there's some trimming to be done in the `submission_manager/models.py` now, particularly the `emit_status()` method that is no longer used.

auacm/app/modules/submission_manager/judge.py
WilliamHester: Yes, but in this case, [it's not possible.](https://www.python.org/dev/peps/pep-0008/#id18)

> Backslashes may still be appropriate at times. For example, long, multiple with -statements cannot use implicit continuation, so backslashes are acceptable

BrandonLMorris:
Awesome, looks fantastic.

PR #74, opened 2016-02-26
Ranklist

### LINKED ISSUE #28: Add Ranklist
View users based off of number of correct submissions or some other criteria.

BrandonLMorris:
Display users ranked by number of problems solved within a time frame. Adds a "Ranking" tab to the navigation bar that directs to the ranking listing, where the time frame can be selected. Rankings are evaluated again upon selecting a different time frame.

This PR also includes some refactoring, namely moving several routes out of app/view.py into user_manager/views.py.

Changes to the database were made: adding a `problem_solved` table that updates when a user solves a problem for the first time.

Fixes #28


BrandonLMorris:
This was a little bit larger than I had originally intended. I think I did a pretty good job separating the work in my commits; it may be easier to roll through those in order than to view all the changes at once.

The SQL commands to create and populate the new `problem_solved` table are as follows:

``` sql
DROP TABLE IF EXISTS problem_solved;
CREATE TABLE problem_solved
(
  username VARCHAR(32) DEFAULT '' NOT NULL,
  pid INT(11) DEFAULT '0' NOT NULL,
  submit_time INT(11) UNSIGNED NOT NULL default 0,
  CONSTRAINT `PRIMARY` PRIMARY KEY (username, pid)
);

INSERT IGNORE INTO problem_solved SELECT username, pid, submit_time FROM submits WHERE result='good';
```

auacm/app/modules/user_manager/views.py
```diff
@@ -0,0 +1,124 @@
+"""User-related routes and methods"""
```
WilliamHester, line 1: I'm actually really glad you moved this here. I was thinking about doing it.

WilliamHester:
Looks great!

PR #79, opened 2016-02-27
Submit API and Recent Submission

BrandonLMorris:
Changes
- Add API routes to obtain submission(s)
  - Submissions can be individually retrieved by specifying the job id, or in batch by specifying the user.
- Judge page lists the 10 most recent submissions of the user (most recent on top), and updates for new submissions.

Notes
- Known issue where the recent submissions won't display if `$scope.username` isn't defined. This happens on refreshing sometimes when the API calls defined in `MainController` aren't called (similar to the issue with problems not listing).

**This pull request builds off of #74. That PR should be merged first**


WilliamHester:
This looks good for the most part. I did find a couple of bugs, though:
1. If you refresh the judge page, sometimes the recent submissions load. Sometimes, they don't.
2. If you log out/switch users from the judge page, the recent submissions do not change.

BrandonLMorris:
1. I actually listed your first point in my initial description of the PR as a known issue. Truth be told, I don't know exactly why the API calls of `MainController` sometimes don't run on refresh. We should probably look into moving those kinds of field to the module level, or maybe providing them through services or something.
2. I didn't try logging in/out, but I'm going to bet that it has to do with the first issue. To my knowledge, logging in/out won't trigger the `$scope.username` field to get reset (at least not in the lower scope of the `JudgeController`)

If you have suggestions, I'll hear them, but I don't think the root cause is with my code, but how we (don't) provide app-level values.

WilliamHester:
I think I was able to fix it relatively easily.
1. Sorry for missing that.
2. It actually does trigger the `$scope.username` field to get reset. [See this](https://docs.angularjs.org/guide/scope).

Also, here is the code that I added to `JudgeController.js` to fix the logging-in/out issue.

``` javascript
$scope.$watch('loggedIn', function(newValue, oldValue) {
    if (newValue) {
        getSubmits();
    } else {
        $scope.submitted = [];
    }
});
```

auacm/app/modules/submission_manager/views.py
```diff
@@ -69,3 +69,44 @@ def submit():
...
+
+    if not submit:
+        return serve_error('Submission with job id ' + str(job_id) +
+                           ' not found', 401)
+
+    return serve_response(submit.to_dict())
+
+
+@app.route('/api/submit/user/<username>')
+@app.route('/api/submit/user/<username>/<int:limit>')
```
WilliamHester: This was way harder to figure out than I thought it would be.

Can we change these to something more apt? I understand if we want to do something like see recent submissions for a user accessible to anyone, but I think we should rename these. Making the route default to the current user would make sense; that's what we'll do most of the time anyway. The use of a query string here would make a lot of sense; it can serve as a sort of filter. Here is what I propose:

For all users: `/api/submissions`

For getting another user's submissions (accessible through `request.args.get('username')`): `/api/submissions?username=<username>`

For getting submissions with a limit (accessible through `request.args.get('limit')`): `/api/submissions?limit=<int:limit>`

For the currently logged-in user (I kind of hate this one, but this is as good as I could come up with): `/api/submissions/mine`

While we're making these changes, I think we should also make a change to the judge route. I wrote it before I really knew what I was doing. The route for it should likely be something like `/api/submissions` but with the method `POST`. I can do this one in a new branch if you'd like.

auacm/app/modules/submission_manager/views.py
```diff
@@ -69,3 +69,44 @@ def submit():
     return serve_response({
         'submissionId': attempt.job
     })
+
+
+@app.route('/api/submit/job/<int:job_id>')
```
WilliamHester: Similar to the note below, I think we should use a different route here to remain RESTful. Since we're accessing a particular submissions, we should be doing something more like `/api/submission/<int:job>`, since we're accessing a specific submission.

auacm/app/modules/submission_manager/views.py
WilliamHester: Alternatively, we could do something like `/api/me/submissions` and possibly even `/api/users/<username>/submissions`.

BrandonLMorris: I agree these need to be changed to query string parameters. Some comments,
- I'm perfectly content leaving out the call to get the current user's submit. It's a shortcut, and I don't think it's an undue burden to ask the client to keep up with the current user. If you really disagree, I'd lean toward `/api/submit/me` (only because `me` is used elsewhere in the API), though I don't love it.
- I don't like putting submissions under the users route. It's very tempting, but I think submits and users are separate actors in the design. After all, we could have users down the road that don't even associate with APT but are still members on the site (which may mean we need to rethink how I did the ranklist).
- I agree that the `submit` route should be changed, though maybe not in this PR. I'd lean to making it `submission` rather than `submissions` just because there's nothing inherently plural about it.

WilliamHester: - Yeah, I kind of hate the `../me` thing. Let's just make the client specify a username.
- I'm fine with leaving it outside the users route. It's not really a property of users, and the query string makes the most sense here.
- We can leave the `submit` route as-is for now. You're right that it's not too related to this pull, but since this pull is making the submissions manager more fully fleshed-out, it kind of makes sense. Also, I still lean towards `submissions` not `submission`, because you are creating something that will be a part of the submissions collection. Wikipedia talks about it more [here](https://en.wikipedia.org/wiki/Representational_state_transfer#Example).

So, can you change the route to `/api/submissions` for this one and `/api/submissions/<int:job>` for the other?

BrandonLMorris:
The corrections were made to the routes to make them much more RESTful. Thanks for bringing it up, I don't really understand REST as much as I should and I completely forgot about query string parameters.

I did _not_ go ahead and change the `api/submit` to `api/submissions`. I really didn't want to dig into the client code.

WilliamHester:
I like most of the changes here. I'm just missing why the `api/submits/<int:job>` was dropped in favor of a query string. Since you are getting a specific submission, it should be in the path, I believe. I feel like it would keep the code cleaner as well, since it seems that you simply made one function into two by adding an if-statement. If you have strong feelings towards the query string, we can leave it, but I think getting an individual submission would best be done through the path.

BrandonLMorris:
Hm, you're right. I think I originally left it out because it could get misinterpreted as the limit (hence the route madness of my first attempt), and then it got lost in the refactoring.

WilliamHester:
Cool. Well once that is resolved, I'll merge it in.

BrandonLMorris:
API design is hard.

WilliamHester:
That it is. I spent about an hour trying to come up with a good solution to it yesterday.

The pull looks good. I'm merging it in.

PR #80, opened 2016-02-29
Replace <select> with Typeahead

WilliamHester:


BrandonLMorris:
Looks great, sorry it took me so long to add this.

PR #89, opened 2016-04-11
Docker

BrandonLMorris:
The app is Dockerized!

Docker allows for a consistent, easily reproducible that can be run on pretty much any platform. There is a bit of a learning curve, but I tried to make it as simple and easy-to-use as possible. Any feedback is appreciated.

I ~~expect~~ hope that this might provide a better way for us to reliably deploy and update, and maybe become our default method of building/shipping the app. This is still new and I don't know all the ins and outs, so I'm definitely looking for feedback. Even if it's not our primary method, it would be very nice if the app supported containerization.

Instructions to build and run are in the `README.md`


BrandonLMorris:
Added docker-machine instructions, should be ready to merge now.

WilliamHester:
Okay, so this looks good to me. I'm all set to merge it. The only comment I have is that we should ideally have a way to persist database changes. We don't want to have something happen where we lose data because the container goes down or because someone accidentally writes the wrong command.

For now, my recommendation is to use this only for development, but as we become more comfortable using it, we can migrate to using it on the server.

PR #88, opened 2016-04-10
Blog tests

BrandonLMorris:
## Changes
- CRUD tests for blog posts
- Bug fixes for blog posts (new ones were breaking)
- Abstract tests to an `AUACMTest` base class

PR #90, opened 2016-04-20
Competition tests

BrandonLMorris:
### Changes
- General smoke tests for competition CRUD-ing
  - **Not** meant to be full-featured tests, just general indications that API endpoints aren't _completely_ broken
- Add `DELETE` endpoint for specific competitions (`/api/competitions/<int:cid>`)
#### Further considerations
- These tests, in reality, barely scratch the surface. Not sure what kind of code coverage we're shooting for, but this is minimal

**Note** This PR includes changes from #88, which should be merged first

### DIFF
--- a/auacm/app/modules/competition_manager/views.py
+++ b/auacm/app/modules/competition_manager/views.py
@@ -202,6 +202,34 @@ def get_competition_data(cid):
     })
 
 
+@app.route('/api/competitions/<int:cid>', methods=['DELETE'])
+def delete_competition(cid):
+    """Delete a competition and all the data associated with it"""
+    competition = (database.session.query(Competition)
+                   .filter_by(cid=cid).first())
+
+    if not competition:
+        return '404: Competition not found', 404
+
+    # Delete the problems from a competition
+    competition_problems = (database.session.query(CompProblem)
+                            .filter_by(cid=cid).all())
+    for competition_problem in competition_problems:
+        database.session.delete(competition_problem)
+
+    # Delete the users/tams from a competition
+    competition_users = (database.session.query(CompUser)
+                         .filter_by(cid=cid).all())
+    for user in competition_users:
+        database.session.delete(user)
+
+    # Delete the competition itself
+    database.session.delete(competition)
+    database.session.commit()
+
+    return 'Successfully deleted', 204
+
+
 @app.route('/api/competitions/<int:cid>/register', methods=['POST'])
 @login_required
 def register_for_competition(cid):
@@ -209,14 +237,14 @@ def register_for_competition(cid):
 
     All the user has to do is submit a post to this url with no form data.
     From their logged-in status, we'll go ahead and add them to the competiton
-    as an individual (team name is default their display name). A 400 error will
-    be returned if the user is already registered for the competition.
+    as an individual (team name is default their display name). A 400 error
+    will be returned if the user is already registered for the competition.
 
     If the user that is submitting this is an admin, they can optionally
-    supply a json array of usernames to register for the competition. Specifying
-    this will not register the admin, but it will register all users that are
-    listed. A 400 error will be returned if any of the users are already
-    registered for the competition.
+    supply a json array of usernames to register for the competition.
+    Specifying this will not register the admin, but it will register all users
+    that are listed. A 400 error will be returned if any of the users are
+    already registered for the competition.
     """
     if database.session.query(Competition).filter(
          Competition.cid == cid).first() is None:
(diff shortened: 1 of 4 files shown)


auacm/app/modules/competition_manager/views.py
```diff
@@ -1,7 +1,6 @@
 from flask import request
 from flask.ext.login import current_user, login_required
 from app import app
-from app.database import session
```
WilliamHester, line 4: What's the motivation to move away from `from app.database import session` to `import app.database as database` then accessing `session` by `database.session`?

auacm/app/modules/competition_manager/views.py
```diff
@@ -116,39 +119,41 @@ def update_competition_data(cid):
...
+    database.session.commit()
     return serve_response(competition.to_dict())
 
 
 @app.route('/api/competitions/<int:cid>')
 def get_competition_data(cid):
-    competition = session.query(Competition).filter(Competition.cid == cid).\
-            first()
+    competition = database.session.query(Competition).filter(
+        Competition.cid == cid).first()
```
WilliamHester, line 136: Shouldn't this be tabbed over one more?

auacm/app/modules/competition_manager/views.py
```diff
@@ -197,57 +202,90 @@ def get_competition_data(cid):
...
 
 
+@app.route('/api/competitions/<int:cid>', methods=['DELETE'])
+def delete_competition(cid):
+    """Delete a competition and all the data associated with it"""
+    competition = (database.session.query(Competition)
+                   .filter_by(cid=cid).first())
+
+    if not competition:
+        return '404: Competition not found', 404
```
WilliamHester, line 212: Why aren't we using `serve_error()` here?

auacm/app/modules/competition_manager/views.py
BrandonLMorris, line 4: The motivation is that it got the test database to work. To be perfectly honest, I'm not 100% clear on the details, but if you do a `from...import...`, it binds the object to a new name in the importing module's namespace. Normally that isn't a problem, but if the imported object gets modified by a third module, the changes won't be reflected.

The `session` variable is initially set to the production database in `database.py`. It's then imported here when the app gets initialized. For the test suite, the `session` object gets reset to point to the test database. Doing a `from...import...` won't catch that and will still point to the production database.

auacm/app/modules/competition_manager/views.py
```diff
@@ -197,57 +202,90 @@ def get_competition_data(cid):
...
+    competition_users = (database.session.query(CompUser)
+                         .filter_by(cid=cid).all())
+    for user in competition_users:
+        database.session.delete(user)
+
+    # Delete the competition itself
+    database.session.delete(competition)
+    database.session.commit()
+
+    return 'Successfully deleted', 204
```
WilliamHester, line 230: Why aren't we using `serve_response()` here?

Also, I have reservations about using the 204 response code. I know that _technically_ it means "The server successfully processed the request and is not returning any content," however from my experience, it's more idiomatic to simply always use 200 for all responses that were successful. I'm open to leaving this as a 204, but it just seems odd.

auacm/app/modules/competition_manager/views.py
BrandonLMorris, line 136: No, [four space hanging indent, nothing on the first line](https://google.github.io/styleguide/pyguide.html?showone=Indentation#Indentation)

auacm/app/modules/competition_manager/views.py
BrandonLMorris, line 230: From my understanding, any 200-type response is an "ok" response. 200 is just generic; a 204 specifically says that there was nothing to return.

auacm/app/modules/competition_manager/views.py
WilliamHester, line 4: Cool. That makes sense. Just making sure that there was some sort of motivation.

auacm/app/modules/competition_manager/views.py
WilliamHester, line 230: After more thought, I think 204 is actually good. I do think that we should modify it so that we use the `serve_respose()` function though rather than simply returning a string if nothing more just to conform to the rest of the codebase. It might take a couple lines of modification to allow different response codes, but it shouldn't be bad.

BrandonLMorris:
Just realized I set this as a PR with `blog-tests` as the base branch. I'll have to close it and reopen it to point to master. My b.

PR #93, opened 2016-04-27
Ui uniformity

WilliamHester:
The long-awaited UI-Uniformity branch. Check it out. The only thing not touched is blog-posts. Things that happened in this branch:
- Create User, Competition, Problem, and Blog Post have ben removed from buttons on their respective pages and added into one '+' button at the top for admins only.
- All tables have been wrapped in a panel and have their proper `<thead>` and `<tbody>` tags.
- General cleanup of HTML
- Fix of weird bug where the search icon was off by 1px
- All of the columns in tables have been aligned using `col-md-x` classes

Let me know what you think.


auacm/app/static/html/competitions.html
```diff
@@ -1,70 +1,81 @@
...
-          <a ng-if="!competition.registered && (isAdmin || !competition.closed)"
-              href="" ng-click="register(competition)">Register</a>
-          <div ng-if="!competition.registered && !isAdmin && competition.closed">Closed</div>
-        </td>
-      </tr>
+  <div class="panel panel-default" ng-show="competitions.upcoming.length > 0">
+    <div class="panel-heading">
+      <h2 class="panel-title">Upcoming</h2>
+    </div>
+    <table c
```
BrandonLMorris: Looks like you forgot the class name here. Makes the table not look right.

BrandonLMorris:
I like the `+` button. However, the create blog form and create competition form look different than creating user and problem.

auacm/app/static/html/competitions.html
WilliamHester: Looks like this was an accident when merging files.

BrandonLMorris:
Me lo gusta mucho. Only a few nits to pick before I merge
- [x] Missing HTML class in Upcoming competition table
- [x] New blog post form does not match new problem/user forms (may be result of merges)
- [ ] [Opinion] Blog posts could be in some kind of well or border
  - If we change the background color, the text becomes unreadable. Maybe should cross that bridge when we get to it.

I mentioned earlier that the new competition form isn't uniform, but I think that's justified given the nature of creating a competition.

WilliamHester:
I'm going to work on the blog post form and new problem/user forms tomorrow after my final. I honestly think that that was just something I forgot, because this branch saw no activity for over a month.

BrandonLMorris:
Haha, what are finals?

WilliamHester:
So, I put the "Create Competition" and  "Create new blog post" forms inside of panels. They all match except for the fact that the "Save" button of the "Create Competition" form is actually in the heading instead of at the bottom. I'm open to moving it, but I feel like it looks better there instead of below the table.

I agree that blog posts should be fixed, but I don't think they need a panel. I'm not sure what exactly would make them look good. I think we should leave that for another pull-request.

auacm/app/static/html/editCompetition.html
```diff
@@ -53,27 +53,25 @@ <h1 class="pull-left">{{ title }}</h1>
...
+
+  <table class="table table-bordered">
+    <thead>
+      <tr>
+        <th class="col-sm-4">Label</th>
+        <th class="col-sm-4">Name</th>
+        <th class="col-sm-4"></th> <!-- This is reserved for the add/remove button -->
+      </tr>
+    </thead>
+    <tbody>
```
BrandonLMorris, line 74: It looks like the top border of the table isn't showing up. Not wrong, just looks a little off.

auacm/app/static/html/editCompetition.html
```diff
@@ -1,18 +1,18 @@
...
+        </button>
+      </div>
     </div>
   </div>
+
+  <h3 class="col-md-12">Metadata</h3>
+
   <div class="panel-body">
 
     <form name="createCompetitionForm" class="form-horizontal">
```
BrandonLMorris: What do you think about putting this part of the form in a well (adding the `well` class is all it takes). It just seems that the subtitle gets mixed in with the fields visually.

WilliamHester: I like the `well`. Added it.

auacm/app/static/html/editCompetition.html
WilliamHester, line 74: I agree. This looks weird, but for some reason, the tables just don't get top borders. How do you feel about changing it to `table-striped` instead of `table-bordered`? I think it looks pretty good.

BrandonLMorris, line 74: Works for me. Most of our other tables are striped anyway.

WilliamHester:
Merge away!

BrandonLMorris:
Cool sauce. We are not on the same schedule.

PR #94, opened 2016-05-02
Competition tests

BrandonLMorris:
Same as #90, but this time set to merge against master.

Also uses `serve_response` and `serve_error`, as discussed in #90


WilliamHester:
Looks good!

PR #97, opened 2016-05-18
Problem comp release

BrandonLMorris:
Problems can now be associated with a competition and hidden from users until that competition begins.

**Changes**
- Getting one problem will check if that problem's `comp_release` has started. This adds an extra database query.
- Getting _all_ problems adds a database query to the competitions table, but only one. Start times are stored and checked in memory to minimize performance hit.
- On the front-end, creating/editing a problem has a new competition release field with a dropdown of all upcoming competitions. If selected and saved, the problem is inaccessible to _all_ users (admins can still view by going to the edit route)
  - **Note:** that this means that previously unhidden problems can be made hidden by associating them with a new competition
  - **Note:** adding a problem to a competition does _not_ automatically hide it

This feature also includes unit tests. Because that should be a thing.

### DIFF
--- a/auacm/app/modules/problem_manager/views.py
+++ b/auacm/app/modules/problem_manager/views.py
@@ -23,10 +23,10 @@ from time import time
 @app.route('/problems/<shortname>/info.pdf', methods=['GET'])
 def get_problem_info(shortname):
     """Serve the PDF description of a problem"""
-    pid = database.session.query(Problem).\
-            options(load_only('pid', 'shortname')).\
-            filter(Problem.shortname == shortname).\
-            first().pid
+    pid = (database.session.query(Problem)
+           .options(load_only('pid', 'shortname'))
+           .filter(Problem.shortname == shortname)
+           .first().pid)
     return serve_info_pdf(str(pid))
 
 
@@ -80,10 +80,10 @@ def get_problems():
     solved_set = set()
 
     if not current_user.is_anonymous:
-        solved = database.session.query(Submission).\
-                filter(Submission.username == current_user.username).\
-                filter(Submission.result == "good").\
-                all()
+        solved = (database.session.query(Submission)
+                  .filter(Submission.username == current_user.username)
+                  .filter(Submission.result == "good")
+                  .all())
         for solve in solved:
             solved_set.add(solve.pid)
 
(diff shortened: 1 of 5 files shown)


auacm/app/static/js/controllers/EditProblemController.js
```diff
@@ -65,6 +75,8 @@ app.controller('EditProblemController', ['$scope', '$route', '$http',
             fd.append('sol_file', $scope.solFile);
         if (typeof prob.appeared !== 'undefined')
             fd.append('appeared_in', prob.appeared);
+        if (typeof prob.comp_release !== 'undefined');
```
WilliamHester: Not sure you meant to add a semicolon here.

BrandonLMorris: You're right. I hate to be that guy, but a frontend test might have caught this...

auacm/app/modules/problem_manager/views.py
```diff
@@ -92,27 +78,32 @@ def get_problems():
...
+        competitions[comp.cid] = comp.start
 
     if not current_user.is_anonymous:
-        solved = database.session.query(Submission).\
-                filter(Submission.username == current_user.username).\
-                filter(Submission.result == "good").\
-                all()
+        solved = (database.session.query(Submission)
+                  .filter(Submission.username == current_user.username)
+                  .filter(Submission.result == "good")
```
WilliamHester: Accidental `"` instead of `'`.

auacm/app/static/js/controllers/EditProblemController.js
WilliamHester: True. That's also why we have code review though.

auacm/app/static/js/controllers/EditProblemController.js
```diff
@@ -65,6 +75,8 @@ app.controller('EditProblemController', ['$scope', '$route', '$http',
...
+            fd.append('comp_release', parseInt(prob.comp_release));
```
WilliamHester, line 79: It appears that we're not actually using the `comp_release` field on the server side, so problems that are added via the dropdown are never actually updated.

BrandonLMorris, line 79: You're right. The API serve them back in a GET, I got confused and though we were setting it as well.

BrandonLMorris:
That should have all the fixes. I went ahead and made it so that admins can see the unreleased problems. Without it, there wasn't any real way to add them to competitions (I'm not exactly sure what I was thinking).

auacm/app/modules/problem_manager/views.py
```diff
@@ -23,49 +13,47 @@
...
     if is_pid(identifier):
         problem = problem.filter(Problem.pid == identifier).first()
     else:
-        problem = problem.\
-                  filter(Problem.shortname == identifier).first()
+        problem = problem.filter(Problem.shortname == identifier).first()
+
+    # Hide unreleased problems to non-admins
+    if problem is None or (current_user.admin != 1 and comp_not_released(
+            problem.Problem.comp_release)
```
WilliamHester: It's not running due to a missed parenthesis here. I'm just going to push that change.

BrandonLMorris: gah

WilliamHester:
Alright, so hiding is working now, but I have one issue. If I create a problem and add it to the competition from the create-a-problem interface, it doesn't actually add it to the competition. If this is intended, then that's alright. Other than that, it seems to be working, so I'm good with merging it.

BrandonLMorris:
_That's not a bug, it's a feature_

It's a known limitation. I can add it here or in another PR. Your call.

WilliamHester:
I'm fine with leaving that functionality right now, because it says "Release with competition," not implying that it will be added to the competition.

If people complain, we can add that in the future.

PR #99, opened 2016-06-15
Fix anonymous users

BrandonLMorris:
Anonymous users couldn't access any of the problems. Now they can.


BrandonLMorris:
Yeah I'm just gonna merge this.

PR #100, opened 2016-08-18
Problem timing

### LINKED ISSUE #15: Assert legitimate time limits for problems.
App should execute judge solutions with judge input and time itself to see how long a program should run. That should be multiplied by some factor, and potentially another factor for different languages. This should be done empirically for all problems.

WilliamHester:
After months of waiting and doing everything but Python development, it's finally here.

It should be good to go, but feel free to make any comments you have about it. Be sure to try it out to see what you think.

One thing that this branch (sadly) does not do is any sort of timing of different programming language for real multipliers. I think that would certainly be a worthwhile project though.

Fixes #15


WilliamHester:
One thing we might want to add to this is some sort of status readout. Right now, it's silent until it's done timing all of the problems. I could add that as either default or via a command-line argument.

BrandonLMorris:
Yeah, we should probably log it somewhere or something. I don't guess we have any logging, do we?

auacm/app/modules/problem_manager/problem_timer.py
```diff
@@ -0,0 +1,121 @@
...
+        self._create_temp_directory()
+        try:
+            self._run()
+        except NoJudgeSolitionsError as e:
+            print(str(e), file=sys.stderr)
+        else:
+            self._remove_temp_files()
+
+
+class NoJudgeSolitionsError(Exception):
```
mcprice30: Should this be "NoJudgeSolutions" ?

WilliamHester: Like, should this be `NoJudgeSolutions` over `NoJudgeSolutionsError`?

WilliamHester:
I added logging, so we should be just about good to go.

auacm/app/modules/problem_manager/problem_timer.py
mcprice30: No, I mean Solutions vs Solitions.

WilliamHester: Good catch.

BrandonLMorris:
The real multipliers, I don't think we were ever planning on ~~putting~~ generating them directly in the project. I was under the impression that we would guess until we got around to empirically testing it out on problems with different complexities. That is to say, the multipliers would be hard-coded and not generated by AUACM

WilliamHester:
I would like to see some performance stats on multiple machines of Java vs Python vs C++, etc. If it happens that the numbers are about the same, then it would make sense to just hardcode the values in, but if we get varying results, then it would make more sense to actually generate those as well. It would be a larger undertaking for a different PR though.

BrandonLMorris:
Why would multiple machines make a difference? They're all going to be judged on the same computer.

I guess I figure that language discrepancy << algorithmic discrepancy. So the multiplier need not be precise, but reasonable. If it becomes an issue, we can handle it then.

auacm/app/modules/problem_manager/problem_timer.py
```diff
@@ -0,0 +1,134 @@
...
+        :raises NoJudgeSolutionError: in the event that no judge solution is
+            found for the given problem.
+        """
+        files = os.listdir(self.path)
+
+        judged_count = 0
+        found_max = 0
+        problem_details = (database.session.query(models.Problem)
+                           .filter(models.Problem.pid == self.problem.pid).first())
+        print('Judging', problem_details.name, '...', file=sys.stderr)
```
BrandonLMorris, line 68: Any particular reason this is printing to `stderr`?

auacm/app/modules/problem_manager/problem_timer.py
```diff
@@ -0,0 +1,134 @@
...
+            sub = MockSubmission(
+                pid=self.problem.pid,
+                file_type=file_type,
+                job=self.problem.pid
+            )
+            submission_path = os.path.join(self.submission_path, str(sub.job))
+            upload = MockUploadFile(source)
+            upload.save(os.path.join(submission_path, fname))
+
+            judgement = judge.Judge(sub.pid, submission_path, upload, 120 if file_type != 'cpp' else 10)
```
BrandonLMorris: Why is `time_limit` 120 for non-C++ and 10 otherwise?

auacm/app/modules/problem_manager/problem_timer.py
```diff
@@ -0,0 +1,134 @@
...
+        print('Judged', problem_details.name, 'with time', found_max)
+
+    def run(self):
+        """Executes the problem timer."""
+        self._create_temp_directory()
+        try:
+            self._run()
+        except NoJudgeSolutionsError as e:
+            print(str(e), file=sys.stderr)
+        else:
```
BrandonLMorris: should this be `finally`?

auacm/app/modules/problem_manager/problem_timer.py
```diff
@@ -0,0 +1,134 @@
...
+        folder. For each file, it creates a mock submission (so that it does
+        not have to store a submission in the database), and a "Mock" upload
+        file. The "Mock" upload file still depends on the computer's
+        filesystem, because the Judge is dependent upon the computer's
+        filesystem.
+
+        Each file is run in serial, and the maximum time of all of the judge
+        submissions, times 1.5 (just to give people some wiggle-room). Then,
+        it is "standardized" by dividing by the judge's TIMEOUT_MULTIPLIER for
+        the given file type.
```
BrandonLMorris, line 57: I don't understand the "standardization". Why divide by the timeout multiplier?

auacm/app/modules/problem_manager/problem_timer.py
```diff
@@ -0,0 +1,134 @@
...
+        OSError will be raised. In that case, you should probably choose a new
+        temp directory as well as delete the old temp directory yourself.
+        """
+        shutil.rmtree(self.submission_path)
+
+    def _run(self):
+        """Times all of the problems in the problem's test folder.
+
+        This will iterate through all of the files in the problem's test
+        folder. For each file, it creates a mock submission (so that it does
```
BrandonLMorris, line 48: Is there any specification of what constitutes a "test folder"? From the look of it, it seems like one or more solutions. Is that correct?

auacm/app/modules/problem_manager/problem_timer.py
WilliamHester: I figured that if it was printed to stderr, it could be ignored easier, and the actual results would be more visible. If you think that that would be pointless, we should instead just print to stdout as we usually do.

auacm/app/modules/problem_manager/problem_timer.py
WilliamHester: Uhh, because I forgot to change it back. The reason I initially changed it is because I have some C++ solutions that kept timing out (I never tried to figure out why though).

auacm/app/modules/problem_manager/problem_timer.py
WilliamHester: Yes, it should.

auacm/app/modules/problem_manager/problem_timer.py
WilliamHester: Our timeout multiplier for Java is 1.5, C++ is 1, and Python is 2.

So, say a raw Python judge solution time is 6 seconds, times our wiggle-room multiplier of 1.5 is 9, then divided by our Python timeout multiplier gives us 4.5, meaning that for Java, the timeout is 6.75 seconds, and for C/C++ it is 4.5 seconds. It should be stored as 4.5 seconds.

auacm/app/modules/problem_manager/problem_timer.py
WilliamHester: There's no definition of a "test folder," though maybe that's something that should be documented. Right now, a test folder is pretty much just a folder inside of the problem's file tree called "test" that contains 0 or more solutions.

BrandonLMorris: Perhaps it should be called a solutions folder. The name seems to predate this, so it's probably out of scope.

auacm/app/modules/problem_manager/problem_timer.py
BrandonLMorris: Oh, gotcha: standardizing across languages. Makes sense.

auacm/app/modules/problem_manager/problem_timer.py
BrandonLMorris: Is it possible that some solutions just take more than 10 seconds? This is an upper bound

BrandonLMorris (commented):
Running this myself, I can see that all the results are output to the console. So, to enter to the database, you would have to manually enter the values. How hard would it be to output a SQL file that could set the problem time limits?

auacm/app/modules/problem_manager/problem_timer.py
BrandonLMorris: Got it. I thought you were trying to report an error, but you're just segregating output. All good.

auacm/app/modules/submission_manager/judge_test.py
```diff
@@ -74,13 +74,15 @@ def createMockSubmission(self, problem, filename):
         )
         file_path = os.path.join(
             data_folder, 'problems', problem, 'solutions', filename)
-        submit_directory = os.path.join(app.config['DATA_FOLDER'], 'submits', str(submit.job))
+        submit_directory = os.path.join(app.config['DATA_FOLDER'], 'submits',
+                str(submit.job))
```
BrandonLMorris: To pick a nit, the bottom line should be indented to the opening paren

``` python
submit_directory = os.path.join(app.config['DATA_FOLDER'], 'submits',
                                str(submit.job))
```

[Source](https://google.github.io/styleguide/pyguide.html?showone=Indentation#Indentation)

auacm/app/modules/submission_manager/judge_test.py
```diff
@@ -74,13 +74,15 @@ def createMockSubmission(self, problem, filename):
...
         if not os.path.exists(submit_directory):
             os.mkdir(submit_directory)
         # Keep track of submit_file so that we can close it.
         self.submit = submit
         self.submit_file = MockUploadFile(file_path)
-        self.judge = judge.Judge(self.submit, self.submit_file, 1)
+        self.judge = judge.Judge(self.submit.pid, submit_directory,
+                self.submit_file, 1)
```
BrandonLMorris: See above

auacm/app/modules/submission_manager/judge.py
```diff
@@ -50,37 +46,39 @@ def allowed_filetype(filename):
...
 
+
 class Judge:
     """A container for a judging instance. It contains the submission, the
     source code that goes along with the submission, and the time limit of the
     problem.
     """
 
-    def __init__(self, submission, uploaded_file, time_limit):
+    def __init__(self, pid, submission_path, uploaded_file, time_limit, on_status=None):
```
BrandonLMorris, line 65: Why is `on_status` a function rather than a bool? Is there something else we might want to do than a websocket emit when a status is changed?

auacm/app/modules/submission_manager/views.py
```diff
@@ -60,11 +60,28 @@ def submit():
 
     attempt.commit_to_session()
 
-    directory = directory_for_submission(attempt)
-    os.mkdir(directory)
-    uploaded_file.save(os.path.join(directory, uploaded_file.filename))
 
-    judge.Judge(attempt, uploaded_file, time_limit).run_threaded()
+    submission_path = os.path.join(
+            app.config['DATA_FOLDER'], 'submits', str(submit.job))
```
BrandonLMorris: this should be `attempt.job`

BrandonLMorris (changes requested):
I'm not sure what it is, but the real-time status seems to be broken. When I submit, the status goes from "Compiling" to blank, but never states correct nor changes the row to green unless I refresh the page.

auacm/app/modules/problem_manager/problem_timer.py
WilliamHester: Yeah, it's certainly possible that some solutions take longer than 10 seconds, though I would wager that 20 seconds is an upper limit for programming competition problems. I'd be fine with putting that as a limit here, standardized of course.

auacm/app/modules/problem_manager/problem_timer.py
WilliamHester: Yeah, I agree on both.

auacm/app/modules/submission_manager/judge.py
WilliamHester: For now, no, there's not anything else that we'd be doing, but it decouples the logic of publishing the status to clients from the judge to the calling function. I think it's cleaner that way. It means that the judge can be run as more of a black box. All logic that actually interacts with the network can be maintained independent of the judge itself.

auacm/app/modules/submission_manager/views.py
WilliamHester: Good catch.

WilliamHester:
As for outputting SQL, that's not necessary, if you check out the `problem_timer.py` file, you can see that if it finds a correct solution with a time greater than that of the previous solution, it updates the problem in the database and commits it. The logging is purely so that someone can see what's going on as it's happening.

As for the error, that's odd, but it shouldn't be caused by this PR. I opened the JS debugger and noticed that it's actually receiving the "good" message from the backend, it's just not doing anything with it. I'll look more into that tomorrow/this weekend to see if I can find the source of that issue.

auacm/app/modules/submission_manager/views.py
```diff
@@ -60,11 +60,28 @@ def submit():
...
+        the submission has a new status.
+        """
+        attempt.update_status(DB_STATUS[status])
+        Flasknado.emit('status', {
+            'submissionId': attempt.job,
+            'problemId': attempt.pid,
+            'username': attempt.username,
+            'submitTime': attempt.submit_time,
+            'testNum': test_number,
+            'status': DB_STATUS[status]
```
BrandonLMorris: I think this is why the real-time judging broke on the front end. `DB_STATUS` doesn't include a `correct` field like the front-end expects. You'll need to either bring back the `EVENT_STATUS` list you deleted, or update the Javascript accordingly

BrandonLMorris (changes requested):
Ah, I didn't realize that it updated the DB. In that case, do we kick off a timing when a new problem gets submitted? I don't think that I saw that here.

auacm/app/modules/submission_manager/views.py
WilliamHester: Good catch. I went ahead and changed the way that this was written to make more sense. The `EVENT_STATUS` variable is back, but it's a map from database status to event status rather than relying on some sort of array indexing.

WilliamHester:
This branch does not kick off the problem timing from problem creation. Since we probably want to do more than just kick it off (we probably want to verify that the submitted judge solution works), I think I will leave that out of this update. As it stands, this is just a utility to generate all of the problems' times, though it is not far off from being able to integrate with uploading a problem (another good reason to have the update_status function a callback rather than bound to the judge itself–you could do something like call back with whether or not a judge solution works rather than creating a submission for that user).

BrandonLMorris:
This is getting stale. I can make the required changes myself if necessary. The site is 70% live and this is crucial to being feature-complete

WilliamHester:
I think this PR should be considered done and merged in. After that, it would make sense to make another one with support for kicking off problem timing when a problem is created.

BrandonLMorris:
You're right. I didn't realize/forgot that status updates had been corrected.

Issue #78, opened 2016-02-27
List Team Members on Scoreboard


BrandonLMorris:
The scoreboard should have the option to list the members of a team right on the scoreboard. Special care will need to be taken so that individuals are not listed twice.

mcprice30:
What is the case where individuals are listed twice? When the team name is the same as one of the participants?

BrandonLMorris:
Yes, "individuals" are really teams consisting of one person with the team name the same as the display name.

mcprice30:
What should we do if, for example, there is a team named "Bob" with Adam, Bob, and Charlie as Members? Should we display Bob as a member in this case?

BrandonLMorris:
That seems appropriate to me. The only time we should hide member(s) is if there's one person on the team, and the team name is the same as the only member.

Btw, are you planning on working on this? If so, go ahead and assign yourself so someone else doesn't start working on it too.

mcprice30:
Sure, I will take a gander at it!

WilliamHester:
As Brandon said, yes.

Currently, we're making the assumption that if a user is the only user on their team and that team name has the same name as their display name, then it's an individual. The API is actually not even concerned with individual teams. Check out the output if you go to [/api/competitions/11/teams](auacm.com/api/competitions/11/teams). You will notice that all teams are listed and contain an array of their members. The client is actually what determines whether or not a team is an individual or a team. A true team is one that has multiple members, so we should display what those members are.

The reasoning for the fake "team" to accommodate an individual is due to the database design. We could open an issue to redesign the database; it would not be a bad one, as I think there are several areas where we could improve, but it would be a fairly lengthy task.

mcprice30:
Right now, it appears as though the /api/competitions/<competition_id>/teams endpoint is never called, and that the teams are returned in the /api/competitions/<competition_id> endpoint. Is there a reason for this? If not, would it be a good idea to make the /api/competitions/<competition_id> no longer return a list of teams? Or, would it be better to integrate the data returned by the /teams endpoint into the generic competition endpoint?

WilliamHester:
[...]

PR #102, opened 2016-08-19
[scoreboard] Add team members to scoreboard

mcprice30:
Teams containing multiple members or a single member with a different name than the team will have the team members listed on the scoreboard.

Addresses #78

### DIFF
--- a/auacm/app/modules/competition_manager/views.py
+++ b/auacm/app/modules/competition_manager/views.py
@@ -158,10 +158,16 @@ def get_competition_data(cid):
     scoreboard = list()
 
     team_users = dict()
+    team_display_names = dict()
+
     for user in comp_users:
         if not user.team in team_users:
             team_users[user.team] = list()
+            team_display_names[user.team] = list()
         team_users[user.team].append(user.username)
+        team_display_names[user.team].append(database.session.query(User)
+                                            .filter(User.username == user.username)
+                                            .first().display)
 
     for team in team_users:
         team_problems = dict()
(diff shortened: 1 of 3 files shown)


auacm/app/modules/competition_manager/views.py
```diff
@@ -158,10 +158,16 @@ def get_competition_data(cid):
...
+    team_display_names = dict()
+
     for user in comp_users:
         if not user.team in team_users:
             team_users[user.team] = list()
+            team_display_names[user.team] = list()
         team_users[user.team].append(user.username)
+        team_display_names[user.team].append(database.session.query(User)
+                                            .filter(User.username == user.username)
+                                            .first().display)
```
BrandonLMorris: This creates a new query for every user in a competition. I don't want to prematurely optimize, but I would be interested if it would be faster to get all the display names (for users in the competition) in one query and stick it in a dictionary for lookup. This may be moot if SQLAlchemy does any kind of caching. I'll do a quick performance test and see if there's a difference.

BrandonLMorris: Yeah, it looks like there's no caching, and [queries are expensive](https://gist.github.com/BrandonLMorris/3e2ae042911026a9f04ec40f33dbfb98). What do you think @mcprice30 ?

mcprice30: Fixed.

BrandonLMorris:
Looks gooooood

PR #103, opened 2017-02-11
Update docs

mcprice30:
Maybe not perfect, but lots of improvements to our documentation.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -57,7 +57,7 @@ Those logged in as an administrator can perform additional actions:
 |`sample_cases`|`Sample Case[]`|A list of samples cases, as described below|
 |`shortname`|`String`|A unique string with no spaces used to identify the problem|
 
-The `sample_cases` field consists of an array of [Sample Case](#the-sample-case-object) object.
+The `sample_cases` field consists of an array of [Sample Case](#the-sample-case-object) objects.
 
 
 __Example:__
--- a/docs/api.md
+++ b/docs/api.md
@@ -7,7 +7,7 @@ method arguments and return values.
 ### Table of Contents
 
 1. [Problem Management](#problem-management)
-2. [Competition Management](#submission-management)
+2. [Competition Management](#competition-management)
 3. [Submission Management](#submission-management)
 4. [Blog Management](#submission-management)
 5. [User Management](#user-management)
@@ -227,21 +227,39 @@ __Example uses:__
 
 ## Competition Management
 
+## Competition Management Objects
+
+### The Competition Object
+
 Competitions are always represented in JSON with the following format:
 
-  * __cid:__ The integer competition ID
-  * __closed:__ `true` if administrators are the only users that can register
-   participants, `false` otherwise
-  * __length:__ The length of the competition, in seconds
-  * __name:__ The string name of this competition
-  * __registered:__ `true` if the current user is registered for this
-   comptition, `false` otherwise
-  * __startTime:__ The time the competition starts, in seconds since the Unix
-   epoch
-  * __compProblems:__ An array of all the problems in the competition
+| Name | Type | Description |
+| --- | --- | --- |
+|`cid`|`int`|The integer competition ID|
+|`closed`|`boolean`|`true` if administrators are the only users that can register participants, `false` otherwise|
+|`length`|`int`|The length of the competition, in seconds|
+|`name`|`String`|The name of this competition|
+|`registered`|`boolean`|`true` if the current user is registered for this competition, `false` otherwise|
+|`startTime`|`int`|The time the competition starts, in seconds since the Unix epoch|
+|`compProblems`|`Competition Problems`|An array of all the problems in the competition|
 
 
-The `compProblems` object is formatted as follows:
+### The Competition Problems Object
+
+The Competition Problems object represents the collection of problems that appear
+in a particular competition. It consists of a mapping of problem letters
+(i.e. "A" for the first problem, "B" for the second, etc.) to a Competition Problem
+object, described below:
+
+__Competition Problem:__
+
+| Name | Type | Description |
+| --- | --- | --- |
+|`name`|`String`|The full name of the problem|
+|`pid`|`int`|An integer uniquely identifying the problem|
+|`shortname`|`String`|A string that can be used to uniquely identify the problem|
+
+__Example Competition Problems Object:__
 
 ```json
 "compProblems": {
@@ -257,6 +275,17 @@ The `compProblems` object is formatted as follows:
   }
 }
 ```
+
+In the above example, there are two problems in the competition. The first, problem
+"A", is named "Islands in the Data Stream", with problem ID 23 and the shortname
+"islands", while the second, problem "B" is named "Von Neumann's Fly", and has
+problem ID 63 with the shortname "vonneumann".
+
+
+### The Team Object
+
+The `compProblems` object is formatted as follows:
+
 
 The API will also accept a Web Socket message `system_time` with no data. The
 API will reply with another Web Socket message with the same name and the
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -7,102 +7,167 @@ method arguments and return values.
...
+5. [Delete a Problem](#delete-a-problem)
+
+
+## Problem Management Objects
+
+### The Problem Object
+
+| Name | Type | Description |
+| --- | --- | --- |
+|`added`|`int`|The UTC timestamp indicating when this problem was uploaded|
```
WilliamHester: In seconds

docs/api.md
```diff
@@ -7,102 +7,167 @@ method arguments and return values.
...
+|`appeared`|`String`|The contest this problem originally appeared in|
+|`comp_release`|`int`|The competition `cid` for the competition this was used in|
+|`description`|`String`|The problem's description|
```
WilliamHester: Formatted in markdown

docs/api.md
```diff
@@ -7,102 +7,167 @@ method arguments and return values.
...
+|`difficulty`|`int`|The difficulty of the problem from 0 (easiest) to 100 (hardest)|
+|`input_desc`|`String`|A description of the problem's input|
+|`name`|`String`|The name of the problem|
+|`output_desc`|`String`|A description of the problem's output|
+|`pid`|`int`|The problem's unique id.|
+|`sample_cases`|`Sample Case[]`|A list of samples cases, as described below|
```
WilliamHester: I prefer `SampleCase[]` over `Sample Case[]`; it makes it seem more like a type.

docs/api.md
```diff
@@ -7,102 +7,167 @@ method arguments and return values.
...
+|`shortname`|`String`|A unique string with no spaces used to identify the problem|
+
+The `sample_cases` field consists of an array of [Sample Case](#the-sample-case-object) objects.
```
WilliamHester: See above, but actually I don't understand why this line is necessary at all. It's already enumerated in the table above.

docs/api.md
```diff
@@ -7,102 +7,167 @@ method arguments and return values.
...
+
+
+__Example:__
 
 ```json
-[
-  {
+{
```
WilliamHester: Several of the fields here are either null or empty. For sample data, it makes sense to fill it all out so that the reader of the documentation can actually understand what it will look like.

docs/api.md
```diff
@@ -7,102 +7,167 @@ method arguments and return values.
...
+
+| Name | Type | Description |
+| --- | --- | --- |
+|`input`|`String`|The input fed to standard in for a test run|
+|`output`|`String`|The output that would be produced to standard out by a correct program for that test run|
+
+__Example:__
+
+```json
+{
```
WilliamHester, line 108: Minor thing, but if you can find a sample with more data, it may be better. Totally up to you.

docs/api.md
```diff
@@ -7,102 +7,167 @@ method arguments and return values.
...
+
+### Get Data on all problems
+
+__URL:__ `/api/problems`
+
+__Method:__ `GET`
+
+__Returns:__ a JSON array of all the publicly available data on all the problems
+in the database.
+Specifically, the method returns an array of [problem objects](#the-problem-object)
```
WilliamHester: Should be "Problem objects", since Problem is a type.

docs/api.md
```diff
@@ -7,102 +7,167 @@ method arguments and return values.
...
-  * __Short Name__
-  * __Competition Appearance__
-  * __Difficulty Rating__
-  * __AUACM Competition Release__
-  * __Description__
-  * __Input Description__
-  * __Output Description__
-  * __Sample Cases Array__
+__Returns:__ detailed data on a specific problem as a JSON object.
+Specifically, the method returns a [problem object](#the-problem-object).
```
WilliamHester: Capitalize "Problem."

docs/api.md
```diff
@@ -7,102 +7,167 @@ method arguments and return values.
...
-  * __competition appearance__
-  * __difficulty rating__
-  * __auacm competition release__
-  * __date added__
-  * __solved by user__
-  * __problem url__
-
-of all the problems.
+### Create a New Problem
+***This method requires being logged in as an administrator***
```
WilliamHester: I'd prefer "This endpoint..." over "This method...".

Also, consider grouping all of the admin-only endpoints into one section.

docs/api.md
```diff
@@ -7,102 +7,167 @@ method arguments and return values.
...
+
+__URL:__ `/api/problems`
+
+__Method:__ `POST`
+
+__Form Data:__
+
+| Title | Required | Form Name | Description|
```
WilliamHester: Consider adding a type here, just for clarity

docs/api.md
```diff
@@ -7,102 +7,167 @@ method arguments and return values.
...
+| --- | --- | --- | --- |
+| Problem Name | __Required__ | `name` | Full problem name (<= 32 characters) |
+| Description | __Required__ | `description` | Problem background story |
+| Input Description | __Required__ | `input_desc` | Description of input values |
+| Output Description | __Required__ | `output_desc` | Description of output values |
+| Sample Case(s) | __Required__ | `cases` | JSON array of sample input/output (see below) |
+| Time Limit | Optional | `time_limit` | Max execution time (in seconds) |
+| Difficulty | Optional | `difficulty` | 0-100 difficulty rating |
+| Appeared In | Optional | `appeared_in` | String name of original competition |
+| Input Files | __Required__ | `in_file` | Zipped (.zip) directory of all input files |
```
WilliamHester: Maybe add a note about how this zip should be structured?

E.g. "Zipped (.zip) directory of all input files ([Format](#zip-format))"

docs/api.md
```diff
@@ -7,102 +7,167 @@ method arguments and return values.
...
+| Output Files | __Required__ | `out_file` | Zipped (.zip) directory of all output files |
```
WilliamHester: Same

docs/api.md
```diff
@@ -7,102 +7,167 @@ method arguments and return values.
...
+| Solution File | __Required__ | `sol_file` | Solution program (not zipped) |
+
+The `cases` field consists of a JSON array of [Sample Case](#the-sample-case-object) objects.
```
WilliamHester: `SampleCase` over `Sample Case`

docs/api.md
```diff
@@ -7,102 +7,167 @@ method arguments and return values.
...
+
+__Returns:__ If any of the required fields are not supplied, the API will return with status
```
WilliamHester: Very minor, but I'd phrase it so that you get the normal result first, followed by the exception case.

E.g.: 
__Returns:__ A [Problem](#the-problem-object) if successful. If unsuccessful, an error response and an HTTP code 400.

docs/api.md
```diff
@@ -7,102 +7,167 @@ method arguments and return values.
...
+code 400. If successful, the API will return a [JSON object representation](#the-problem-object)
```
WilliamHester: You probably don't need to specify JSON, since this should be documented as a JSON API.

docs/api.md
```diff
@@ -129,9 +194,11 @@ __Form Data__
 | Output Files | Optional | `out_file` | Zipped (.zip) directory of judge output files (must be titled `out.zip`) |
 | Judge Solution | Optional | `sol_file` | Judge solution program (do not zip) |
 
+The `cases` field consists of a JSON array of [Sample Case](#the-sample-case-object) objects.
```
WilliamHester: Ditto with Sample Cases

WilliamHester (changes requested):

WilliamHester (approved):

PR #105, opened 2017-03-07
Fix issue with problems not uploading

Mac-Genius:


auacm/app/modules/problem_manager/views.py
```diff
@@ -139,7 +139,13 @@ def create_problem():
             problem.difficulty = request.form['difficulty']
         if 'appeared_in' in request.form:
             problem.appeared = request.form['appeared_in']
-        problem.comp_release = request.form['comp_release'] or None
+        if 'comp_release' in request.form:
```
WilliamHester: Can we make this more concise? Something more like

```py
if 'comp_release' not in request.form or int(request.form['comp_release']) < 1:
    problem.comp_release = None
else:
    problem.comp_release = request.form['comp_release']
```

Mac-Genius: Yes

WilliamHester (approved):

PR #109, opened 2017-03-23
Fix Markdown rendering breaking when shortening the Markdown string.

Mac-Genius:


auacm/angular/src/app/blog-list/blog-list.component.html
```diff
@@ -23,7 +23,7 @@
           </h4>
         </div>
       </div>
-      <div class="text-left" [innerHTML]="converter.makeHtml(post.expanded ? post.body : post.body | limitWords : 200)"></div>
+      <div class="text-left" [innerHTML]="post.expanded ? converter.makeHtml(post.body) : (post.body | limitWords : 200)"></div>
```
mcprice30: Could you please either re-work `limitWords` to take HTML, rather than the raw markdown, or rename `limitWords` to indicate that this converts the markdown to HTML? The current approach isn't very readable.

auacm/angular/src/app/blog-list/blog-list.component.ts
```diff
@@ -38,7 +38,7 @@ export class BlogListComponent implements OnInit {
       const tempPipe = new LimitWordsPipe();
       this.blogPosts = blogs;
       for (let i = 0; i < this.blogPosts.length; i++) {
-        if (tempPipe.transform(this.blogPosts[i].body, 200).trim().length < this.blogPosts[i].body.trim().length) {
+        if (tempPipe.transform(this.blogPosts[i].body, 200).trim().length < this.converter.makeHtml(this.blogPosts[i].body.trim()).length) {
```
mcprice30: Can we make this 200 be a constant or something? This is a bit of magic number, and this way the HTML and component won't be out of sync.

auacm/angular/src/app/pipes/limit-words.pipe.ts
```diff
@@ -7,18 +8,107 @@ import { Pipe, PipeTransform } from '@angular/core';
...
+    const parsedHtml = document.createElement( 'html' );
+    parsedHtml.innerHTML = htmlWords;
+
+    const tempChildren = parsedHtml.children[1].childNodes;
+    const children = [];
+    for (let i = 0; i < tempChildren.length; i++) {
+      children.push(tempChildren[i]);
+    }
+
+    // let finalHtml = document.createElement( 'html' );
```
mcprice30: Kill this comment line please.

auacm/angular/src/app/pipes/limit-words.pipe.ts
```diff
@@ -7,18 +8,107 @@ import { Pipe, PipeTransform } from '@angular/core';
...
-    for (let i = 0; i < amount && i < allWords.length; i++) {
-      finalString += allWords[i] + ' ';
-      count++;
-    };
-    if (count === amount) {
-      return finalString.trim() + ' ...';
+    let size = 0;
+    for (let i = 0; i < children.length; i++) {
+      const element = this.getStringForElement(children[i], '', size, amount);
+      finalString += element.value;
```
mcprice30: Is element.size unrelated to element.value? In this case it seems strange to be doing `+=` at one point, but doing a regular assignment to the other one.

I think a more readable approach would be:
```
let element = new CustomString('', 0);
for (let i = 0; i < children.length; i++) {
  element = this.getStringForElement(children[i],  element.value, element.size, amount);
}
return element.value;
```

I'm not 100% sure, but I think that should do the same thing, and to me it makes more sense.

mcprice30: And maybe rename `element` to `finalElement`.

auacm/angular/src/app/pipes/limit-words.pipe.ts
```diff
@@ -7,18 +8,107 @@ import { Pipe, PipeTransform } from '@angular/core';
...
+  /**
+   * Recursively searches through an HTML element and tries to shorten it to the maxSize.
+   *
+   * @param htmlElement - the element to shorten
+   * @param currentString - the current string that has been shortened
+   * @param currentSize - the size of the currentString
+   * @param maxSize - the maximum amount of words that can be within the currentString
+   * @returns a CustomString containing the shortened element as a string and the size of the element
+   */
+  getStringForElement(htmlElement: any, currentString: string, currentSize: number, maxSize: number): CustomString {
```
mcprice30: This method is pretty long. It might be a bit more readable if you broke this into a few helper methods.

mcprice30 (commented):
Looks pretty good!

Just a few minor changes I would recommend.

mcprice30 (approved):
Looks good to me!

PR #111, opened 2017-03-24
Fix 'Continue reading' button appearing on blogposts that have not be…

Mac-Genius:


mcprice30 (commented):
My understanding is that the separate handling of the .trim() is to handle paragraph breaks. If that is the case, then this works for me.

Mac-Genius:
Yes, the separate handling with trim ensures that whitespace does not count towards the running total of words. In the second commit, I removed a trim that was removing important whitespace such as spaces and line breaks.

mcprice30 (approved):
Ok, then.

PR #116, opened 2017-04-05
John/pylint

Mac-Genius:
Brings the backend up to the Pylint standard (Which I believe is PEP 8).

### DIFF
--- a/auacm/app/database.py
+++ b/auacm/app/database.py
@@ -1,17 +1,18 @@
+# pylint: disable=I0011,C0103
 '''Creates the database.'''
 from sqlalchemy import create_engine
 from sqlalchemy.orm import Session
 from sqlalchemy.ext.automap import automap_base
 
 # Create global database variables
-DATABASE_BASE = automap_base()
-DATABASE_ENGINE = create_engine('mysql+pymysql://acm@localhost/acm?charset=utf8')
-DATABASE_CONNECTION = DATABASE_ENGINE.connect()
-DATABASE_SESSION = Session(DATABASE_ENGINE)
+database_base = automap_base()
+database_engine = create_engine('mysql+pymysql://acm@localhost/acm?charset=utf8')
+database_connection = database_engine.connect()
+database_session = Session(database_engine)
 
 def commit_to_session(base):
     '''Add an object to the session and refresh the session.'''
-    DATABASE_SESSION.add(base)
-    DATABASE_SESSION.flush()
-    DATABASE_SESSION.commit()
-    DATABASE_SESSION.refresh(base)
+    database_session.add(base)
+    database_session.flush()
+    database_session.commit()
+    database_session.refresh(base)
--- a/auacm/app/database.py
+++ b/auacm/app/database.py
@@ -1,5 +1,5 @@
 # pylint: disable=I0011,C0103
-'''Creates the database.'''
+"""Creates the database."""
 from sqlalchemy import create_engine
 from sqlalchemy.orm import Session
 from sqlalchemy.ext.automap import automap_base
@@ -11,7 +11,7 @@ database_connection = database_engine.connect()
 database_session = Session(database_engine)
 
 def commit_to_session(base):
-    '''Add an object to the session and refresh the session.'''
+    """Add an object to the session and refresh the session."""
     database_session.add(base)
     database_session.flush()
     database_session.commit()
--- a/auacm/app/database.py
+++ b/auacm/app/database.py
@@ -1,18 +1,17 @@
-'''Database handlers.'''
+'''Creates the database.'''
 from sqlalchemy import create_engine
 from sqlalchemy.orm import Session
 from sqlalchemy.ext.automap import automap_base
 
 # Create global database variables
-Base = automap_base()
-engine = create_engine('mysql+pymysql://acm@localhost/acm?charset=utf8')
-connection = engine.connect()
-session = Session(engine)
+DATABASE_BASE = automap_base()
+DATABASE_ENGINE = create_engine('mysql+pymysql://acm@localhost/acm?charset=utf8')
+DATABASE_CONNECTION = DATABASE_ENGINE.connect()
+DATABASE_SESSION = Session(DATABASE_ENGINE)
 
-# Import all the ORM classes
-from app.modules.user_manager.models import User
-from app.modules.submission_manager.models import Submission, MockSubmission
-from app.modules.competition_manager.models import Competition, CompUser, CompProblem
-from app.modules.problem_manager.models import Problem, ProblemData, SampleCase
-from app.modules.blog_manager.models import BlogPost
-Base.prepare(engine, reflect=True)
+def commit_to_session(base):
+    '''Add an object to the session and refresh the session.'''
+    DATABASE_SESSION.add(base)
+    DATABASE_SESSION.flush()
+    DATABASE_SESSION.commit()
+    DATABASE_SESSION.refresh(base)
--- a/auacm/app/database.py
+++ b/auacm/app/database.py
@@ -6,7 +6,8 @@ from sqlalchemy.ext.automap import automap_base
 
 # Create global database variables
 database_base = automap_base()
-database_engine = create_engine('mysql+pymysql://acm@localhost/acm?charset=utf8')
+database_engine = create_engine(
+    'mysql+pymysql://acm@localhost/acm?charset=utf8')
 database_connection = database_engine.connect()
 database_session = Session(database_engine)
 
--- a/auacm/app/modules/__init__.py
+++ b/auacm/app/modules/__init__.py
@@ -1,8 +1,9 @@
-'''Contains our backend global variables'''
+# pylint: disable=I0011,C0103
+"""Contains our backend global variables"""
 from flask import Flask
 
 # Create the God-objects to be used everywhere
-APP = Flask(__name__, template_folder='../templates',\
-	static_folder='../static')
-APP.config.from_pyfile('../config.py')
-TEST_APP = APP.test_client()
+app = Flask(__name__, template_folder='../templates',
+				  	     static_folder='../static')
+app.config.from_pyfile('../config.py')
+test_app = app.test_client()
--- a/auacm/app/modules/__init__.py
+++ b/auacm/app/modules/__init__.py
@@ -1,0 +1,8 @@
+'''Contains our backend global variables'''
+from flask import Flask
+
+# Create the God-objects to be used everywhere
+APP = Flask(__name__, template_folder='../templates',\
+	static_folder='../static')
+APP.config.from_pyfile('../config.py')
+TEST_APP = APP.test_client()
--- a/auacm/app/modules/blog_manager/blog_test.py
+++ b/auacm/app/modules/blog_manager/blog_test.py
@@ -1,19 +1,17 @@
 """Test for AUACM blog manager """
 
-import unittest, json
+import json
 
-from app import app, test_app
-from app.util import AUACMTest
-from app.modules.blog_manager.models import BlogPost
-import app.database as database
-
-session = database.session
+from ...modules import TEST_APP
+from ...util import AUACMTest
+from ..blog_manager.models import BlogPost
+from ...database import DATABASE_SESSION
 
 
 class AUACMBlogTests(AUACMTest):
     """Test cases for the AUACM blog manager"""
 
-    def testCreate(self):
+    def test_create(self):
         """Test creating a new blog post"""
         post = {
             'title': 'Test Post',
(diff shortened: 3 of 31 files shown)


auacm/app/database.py
```diff
@@ -1,18 +1,17 @@
...
 from sqlalchemy import create_engine
 from sqlalchemy.orm import Session
 from sqlalchemy.ext.automap import automap_base
 
 # Create global database variables
-Base = automap_base()
-engine = create_engine('mysql+pymysql://acm@localhost/acm?charset=utf8')
-connection = engine.connect()
-session = Session(engine)
+DATABASE_BASE = automap_base()
```
BrandonLMorris: I think I'm ok with this, but I'd like a second opinion ( @mcprice30 @WilliamHester ).

My guess is that you uppercased these to satisfy pylint. But uppercasing like this implies that these values are constants, which they're not: they're global objects.

I know global objects are generally taboo for good reasons, but I'm not sure how we would reasonably go about avoiding them for at least some things like `app` and `session`. My only idea is a singleton that creates/returns the object on a getter method.

So I see three options, and I'm not sure which is best:
1. Keep it as is, and just know that the caps doesn't mean it's a constant
1. Disable pylint for these specific object via comments
1. Reorganize it so that we don't have explicitly global objects

Thoughts?

auacm/app/modules/blog_manager/blog_test.py
```diff
@@ -35,59 +33,57 @@ def testCreate(self):
...
-            BlogPost.id == post_response['id']).first()
-        session.delete(post)
-        session.commit()
+        DATABASE_SESSION.flush()
+        post = DATABASE_SESSION.query(BlogPost).filter(
+            BlogPost.post_id == post_response['id']).first()
+        DATABASE_SESSION.delete(post)
+        DATABASE_SESSION.commit()
 
-    def testGetAll(self):
```
BrandonLMorris: Correct me if this has changed, but when this code was written the canonical convention was to use camelCase for unit test names

Mac-Genius: I believe that is correct, but I feel like camel case is not supposed to be used at all in Python. I think we should use the standard `test_` at the beginning of the test method.

auacm/app/database.py
mcprice30: Not really any thoughts, but I don't think 3 is really going to be feasible. Between 1 and 2, I don't have an opinion.

auacm/app/modules/__init__.py
```diff
@@ -0,0 +1,8 @@
+'''Contains our backend global variables'''
```
mcprice30: Nit: can we choose to use either single or double quotes for docstrings? Mixing them feels odd.

auacm/app/modules/blog_manager/blog_test.py
```diff
@@ -35,59 +33,57 @@ def testCreate(self):
...
         self._assert_posts_equal(post_response, post)
 
-        session.delete(post)
-        session.commit()
+        DATABASE_SESSION.delete(post)
+        DATABASE_SESSION.commit()
 
-    def testDelete(self):
+    def test_delete(self):
         """Test deleteing a blog post"""
```
mcprice30: Spelling: deleting, not deleteing (sorry....)

auacm/app/modules/blog_manager/views.py
```diff
@@ -71,20 +83,18 @@ def create_blog_post():
...
                            response_code=400)
 
-    post.commit_to_session(database.session)
-    database.session.refresh(post)
+    commit_to_session(post)
     return serve_response(create_blog_object(post))
 
 
-@app.route('/api/blog/<int:bid>', methods=['DELETE'])
+@APP.route('/api/blog/<int:bid>', methods=['DELETE'])
```
mcprice30: Should we put an `@admin_required` tag on this?

auacm/app/modules/competition_manager/views.py
```diff
@@ -172,75 +178,82 @@ def get_competition_data(cid):
...
 
-    return serve_response({
-        'competition': competition.to_dict(),
-        'compProblems': comp_problems,
-        'teams': scoreboard
-    })
-
-
-@app.route('/api/competitions/<int:cid>', methods=['DELETE'])
+@APP.route('/api/competitions/<int:cid>', methods=['DELETE'])
```
mcprice30: Do we need an `@admin_required` tag on this?

auacm/app/modules/profile_manager/views.py
```diff
@@ -1,49 +1,53 @@
...
 
-@app.route('/api/profile/image/<username>', methods=['GET'])
+@APP.route('/api/profile/image/<username>', methods=['GET'])
 def get_profile_image(username='tester'):
     """
     Return a user's profile picture.
     """
 
-    imagefile = [filename for filename in os.listdir(join(app.config['DATA_FOLDER'],
+    imagefile = [filename for filename in os.listdir(join(APP.config['DATA_FOLDER'],\
```
mcprice30: Can we replace this with `IMAGE_DIR`?

auacm/app/modules/submission_manager/judge.py
```diff
@@ -195,16 +196,16 @@ def _execute_submission(self):
...
-                          self.sub_output_path, out_file)) as sub_result:
+                         self.sub_output_path, out_file)) as sub_result:
                     correct_lines = correct.readlines()
                     submission_lines = sub_result.readlines()
                     if len(submission_lines) != len(correct_lines):
                         return WRONG_ANSWER, test_number, max_time
 
-                    for gl, sl in zip(correct_lines, submission_lines):
-                        if gl.rstrip('\r\n') != sl.rstrip('\r\n'):
+                    for key, value in zip(correct_lines, submission_lines):
```
mcprice30: I feel like `key` and `value` are not very descriptive. Maybe `judge_line` / `expected_line` and `submission_line` / `actual_line` or something to that effect.

pylintrc
```diff
@@ -0,0 +1,407 @@
+[MASTER]
+
+# Specify a configuration file.
```
mcprice30: Can we kill entries like this that don't actually configure anything?

pylintrc
```diff
@@ -0,0 +1,407 @@
...
+#rcfile=
+
+# Python code to execute, usually for sys.path manipulation such as
+# pygtk.require().
+#init-hook=
+
+# Add files or directories to the blacklist. They should be base names, not
+# paths.
+ignore=CVS
```
mcprice30: I don't know for sure, but it seems like this is not necessary.

pylintrc
```diff
@@ -0,0 +1,407 @@
...
+[ELIF]
+
+# Maximum number of nested blocks for function / method body
+max-nested-blocks=5
+
+
+[FORMAT]
+
+# Maximum number of characters on a single line.
+max-line-length=100
```
mcprice30: `docs/contributing.md` says "Python and JavaScript lines should not exceed 80 characters in length. Should we switch the limit to 100, or bring this down to 80?

pylintrc
```diff
@@ -0,0 +1,407 @@
...
+
+# List of member names, which should be excluded from the protected access
+# warning.
+exclude-protected=_asdict,_fields,_replace,_source,_make
+
+
+[DESIGN]
+
+# Maximum number of arguments for function / method
+max-args=10
```
mcprice30: I know this is minor, and it (hopefully) won't affect anything right now, but can we please bump this down a bit. Allowing 10 arguments still seems like a lot to me. I know Dr. Umphress said  5 +/- 2 is about where people run in to trouble, so I would advocate for 7 arguments at the most.

mcprice30 (commented):
Overall, looks good! Most of what I have are nitpicks, questions about the pylintrc, or things that I saw during the review that weren't really part of the change itself.

auacm/app/modules/blog_manager/blog_test.py
Mac-Genius: really

mcprice30: I even apologized in the comment dude.

Mac-Genius: I know haha. It's all good

auacm/app/modules/blog_manager/views.py
Mac-Genius: Probably

auacm/app/modules/profile_manager/views.py
Mac-Genius: Sure

auacm/app/database.py
WilliamHester: I'm not a fan of keeping it as-is. Uppercasing constants should be reserved for immutable objects that are assigned once.

I'd prefer 3, but I recognize that that's a lot of work for a single refactoring. For now, I'd go with 1 and leave a comment to disable the inspection.

auacm/app/modules/__init__.py
WilliamHester: +1. Double quotes is preferred for docstrings, single quotes for other strings.

auacm/app/modules/__init__.py
```diff
@@ -0,0 +1,8 @@
...
+from flask import Flask
+
+# Create the God-objects to be used everywhere
+APP = Flask(__name__, template_folder='../templates',\
```
WilliamHester: `APP` should be lowercase here, as it's not a constant, related to above comment.

auacm/app/modules/blog_manager/views.py
```diff
@@ -1,15 +1,22 @@
...
-from flask.ext.login import current_user, login_required
-from app import app
-from app.util import serve_response, serve_error, admin_required
-from .models import BlogPost
-from app.modules.user_manager.models import User
+from flask.ext.login import current_user
 from sqlalchemy import desc
-from time import time
-import app.database as database
+from ..user_manager.models import User
```
WilliamHester: Relative imports are discouraged. Prefer explicit paths.

auacm/app/modules/blog_manager/views.py
```diff
@@ -1,15 +1,22 @@
...
+from ...database import DATABASE_SESSION
+from ..blog_manager.models import BlogPost
+from ...modules import APP
+from ...database import commit_to_session
+from ...util import serve_error, serve_response, admin_required
 
 def create_blog_object(post):
-    user = database.session.query(User).filter(User.username==post.username)
+    '''Creates a new blog post.'''
+    user = DATABASE_SESSION.query(User)\
```
WilliamHester: Prefer line wrapping with parentheses, not the backslash

auacm/app/modules/blog_manager/views.py
WilliamHester: Yes

WilliamHester (changes requested):

auacm/app/modules/blog_manager/blog_test.py
BrandonLMorris: Judging from the `unittest` docs, it looks like the standard is more or less a mix: use `snake_case` except in standard methods like `setUp()`. So this is fine

pylintrc
Mac-Genius: Sure. Pylint generates the config file and I used it as is. I'll remove the sections we don't use.

pylintrc
Mac-Genius: I actually bumped it up from 5 I think. We have a few methods that had I think 7 or 8 parameters.

auacm/app/modules/blog_manager/views.py
Mac-Genius: I would love to, but the amount of circular dependencies is too high. The last thing I want to do is disable pylint checking for that.

pylintrc
BrandonLMorris: I think those high arguments are from the models. Some of the tables have a lot of columns, and a constructor needs an argument for each field.

Options are:
1. Leave as is
1. Bump down to 5-7 and explicitly disable those areas with comments

auacm/app/modules/__init__.py
```diff
@@ -1,8 +1,9 @@
...
+"""Contains our backend global variables"""
 from flask import Flask
 
 # Create the God-objects to be used everywhere
-APP = Flask(__name__, template_folder='../templates',\
-	static_folder='../static')
-APP.config.from_pyfile('../config.py')
-TEST_APP = APP.test_client()
+app = Flask(__name__, template_folder='../templates',
+				  	     static_folder='../static')
```
mcprice30: Is this really what pylint said to do? It looks insane.

Mac-Genius: It complains if I adjust it.

WilliamHester (approved):

Mac-Genius:
Ready to merge everyone?

PR #117, opened 2017-04-06
Favicon for the site

julienlegault:
This should allow the favicon to show up on any platform/browser. I quickly whipped the accual image itself in a few minutes, if anyone wants it to be something different just comment and I can make a different version.


mcprice30 (changes requested):
While this looks nice (and certainly better than anything I could do), there are a few issues.

1: At the actual 16x16 favicon scale, it's hard to make out.

2: Much more importantly, we actually can't use the Auburn logo thing without express permission and stuff from the university. There's more details about what is necessary here: http://www.auburn.edu/administration/trademark/ but of note is the statement that "promotional licenses must be obtained by companies or organizations wanting to associate with the University through any use of its trademarks and/or images."

I know that this largely deals with things that are sold, but they are pretty strict about use of the logo, especially when it comes to incorporating it into other images. I just think it would be a good idea to check with them first.


Sorry. :(

julienlegault:
I completely agree, I've made a mock-up of a favicon that might be easier to read at lower resolutions, and does not use any copyrighted material. What do you think of this?
acm

Mac-Genius:
Hey @julienlegault, I'm still not a fan of the latest version of your design. I think that the letters are too thin and may make it difficult to read at 16px x 16px. What would be nice would be a simple icon that could be easily recognizable. It does not even need to have ACM on it, but it would need to have something to do with computers/software.

julienlegault:
graphproblem
graphproblemempty
redandgreen
terminal
I whipped up a few ideas, my favorite of which is the red and green one. I tried to get it to look like a section of the scoreboard from during a competition. Let me know if y'all like any of these.

mcprice30:
I like the red and green one the most, too, but it looks like it could be a little more saturated.

julienlegault:
redandgreenv2
redandgreenv3
Yeah I agree it looks better with more saturation, for the first one I took the colors directly from the site, but those are too light for the favicon.

Mac-Genius:
I like the more saturated one, could you try adding rounded corners? I just want to see what it would look like.

PR #122, opened 2017-09-28
Matt/java backend

mvb0005:
I don't think the protobuf is working.


src/main/java/com/auacm/database/model/Problem.java
```diff
@@ -20,18 +20,21 @@
...
     private String difficulty;
 
+    @Transient
+    private String url;
+
     private long added;
 
     @Column(name = "comp_release")
-    private int competitionId;
+    private Integer competitionId;
```
Mac-Genius, line 29: If you can, leave this as an int.

src/main/java/com/auacm/database/model/Problem.java
```diff
@@ -81,11 +84,11 @@ public void setAdded(long added) {
         this.added = added;
     }
 
-    public int getCompetitionId() {
+    public Integer getCompetitionId() {
```
Mac-Genius, line 87: Use int if you can.

src/main/java/com/auacm/database/model/Problem.java
```diff
@@ -81,11 +84,11 @@ public void setAdded(long added) {
...
         return competitionId;
     }
 
-    public void setCompetitionId(int competitionId) {
+    public void setCompetitionId(Integer competitionId) {
```
Mac-Genius, line 91: Int please.

src/main/java/com/auacm/database/model/ProblemData.java
```diff
@@ -1,16 +1,20 @@
...
 @Table(name = "problem_data")
 public class ProblemData implements Serializable {
+
+    @JsonIgnore
     @Id
     private long pid;
 
     @Column(name = "time_limit")
-    private int timeLimit;
+    private Integer timeLimit;
```
Mac-Genius, line 17: Int here as well.

Mac-Genius (changes requested):
Just change Integer to int. Everything else looks good and the problems populate on the problems page.

mvb0005:
`competitionId` cannot be an int as long as there are null values in the database. I'm not sure how to change the default database from null to something more fitting like 0.

Mac-Genius:
Alright, we can leave it as Integer for now then.

Mac-Genius (approved):

